PR #4, opened 2022-06-24
Fix custom webhooks path not working in development

witoszekdev:
I want to merge this because it allows using custom paths for webhooks. Currently in development only the `/pages/api/webhooks` path is used, which makes it impossible to use other path.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,12 +8,16 @@ const capitalize = (value: string) =>
 
 const dropFileExtension = (filename: string) => path.parse(filename).name;
 
-export const inferWebhooks = async (baseURL: string, webhooksPath: string, generatedGraphQL: any) => {
+export const inferWebhooks = async (
+  baseURL: string,
+  webhooksPath: string,
+  generatedGraphQL: any
+) => {
   let entries;
   if (process.env.NODE_ENV === "production") {
-    entries = await fg(["*.js"], { cwd: webhooksPath });
+    entries = await fg(["*.js"], { cwd: `${__dirname}/${webhooksPath}` });
   } else {
-    entries = await fg(["*.ts"], { cwd: `pages/api/webhooks` });
+    entries = await fg(["*.ts"], { cwd: webhooksPath });
   }
 
   return entries.map(dropFileExtension).map((name: string) => {


witoszekdev:
@magul I thought I can just pass whatever directory webhooks for Saleor are located at to the `webhooksPath` parameter and it will work 😅. If we don't want to support this use case then maybe this function shouldn't accept that parameter at all and just use a default value?

The use case for this is that your app might support webhooks not only for Saleor. In checkout app we have webhooks for payment gateways (Mollie, Adyen) and Saleor. Webhooks for Saleor were put into separate folder `/api/webhooks/saleor/...` so that they could be inferred by the SDK: https://github.com/saleor/saleor-checkout/tree/SALEOR-6984-handle-mollie-refunds/apps/checkout-app/pages/api/webhooks

If we don't have support for custom webhook folder then how would we solve that?

magul:
> function shouldn't accept that parameter at all

The problem here was that without it SDK will not know, where to look for webhooks handlers, so we need to provide path of `manifest.ts` to just append `webhooks/` and look for files there, without that parameter I don't know how to make SDK aware there to look for `webhooks` directory (maybe we could just infer that form stacktrace? TBH don't know).

> The use case for this is that your app might support webhooks not only for Saleor. In checkout app we have webhooks for payment gateways (Mollie, Adyen) and Saleor. Webhooks for Saleor were put into separate folder /api/webhooks/saleor/... so that they could be inferred by the SDK:

That's actually valid problem to be solved, I don't really know how to handle it yet :/ Ultimately you can fallback to manually defining webhooks. If you want to figure out something more fancy just be aware, that:
 * directory that holds webhooks should be defined relatively to `manifest.ts`, not SDK
 * discovery of these files differs between development environment and built assets (so, please, test both situations)

lkostrowski:
@witoszekdev @magul any update on this PR?

witoszekdev:
@lkostrowski unfortunately no, I didn't have time to find a solution to this - but now since the webhook registration from manifest is working in Saleor core I think we should improve this 🙈

lkostrowski:
@witoszekdev I see your solution but I dont understand the problem. Is there any case where we can't use this fixed path?

witoszekdev:
@lkostrowski yes, it's the case described here:

> The use case for this is that your app might support webhooks not only for Saleor. In checkout app we have webhooks for payment gateways (Mollie, Adyen) and Saleor. Webhooks for Saleor were put into separate folder /api/webhooks/saleor/... so that they could be inferred by the SDK:

TLDR - you might want to have other webhooks than Saleor and it messes the inference if they would be in the same folder

lkostrowski:
Okay, sorry, didn't read this through.

I've put this to backlog. I think one day we should be able to configure such things, but we are not there yet. 
For now, you can prefix other webhooks (`mollie/webhooks`) as a workaround. Saleor Checkout is... Saleor, so I think we can allow default name to be reserved for Saleor itself :)

lkostrowski:
Closing - `inferWebhooks` was removed. Feel free to implement such logic on the app level

PR #8, opened 2022-07-14
Add types for event validation middleware

witoszekdev:
I want to merge this, because it allows to use stricter types for accepted `expectedEvent` type in `withSaleorEventMatch` middleware

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -31,7 +31,7 @@ export const withSaleorDomainPresent: Middleware =
   };
 
 export const withSaleorEventMatch =
-  (expectedEvent: string): Middleware =>
+  <E extends string>(expectedEvent: `${Lowercase<E>}`): Middleware =>
   (handler) =>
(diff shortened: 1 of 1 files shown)


krzysztofwolski (commented):
Shouldn't we rather compare if event is one of the types allowed in the GQL API?

magul:
> Shouldn't we rather compare if event is one of the types allowed in the GQL API?

In here, in build phase we can do that (just typing that middleware like I did that in https://github.com/saleor/saleor-app-template/pull/55).

Ultimately we also want to check if handler is handling proper event (based on request's header). We store an distinguish id of an event in filename of that handler, passing it once again (in a little different form) to that callable looks like a repetition, but don't know how to avoid it).

krzysztofwolski:
Makes sense! 

In the future, we'll figure out how to include enums from the schema without creating conflicts with different Saleor versions 👍

PR #9, opened 2022-07-20
Add code formatting

krzysztofwolski:
- added eslint and prettier. Used the same setup as saleor/react-storefront
- added commit hook
- CI check linting and build status for all PRs
- formatted code (and renamed a few variables to make their names camelCase)


.eslintrc
```diff
@@ -0,0 +1,58 @@
...
+    "prettier" // prettier *has* to be the last one, to avoid conflicting rules
+  ],
+  "ignorePatterns": ["pnpm-lock.yaml"],
+  "plugins": ["simple-import-sort", "@typescript-eslint"],
+  "rules": {
+    "quotes": ["error", "double"],
+    "react/react-in-jsx-scope": "off", // next does not require react imports
+    "import/extensions": "off", // file extension not required when importing
+    "react/jsx-filename-extension": "off",
+    "no-restricted-syntax": "off",
```
witoszekdev, line 15: Small suggestion to disallow `for ... in` which could be problematic when iterating over objects - for ... in iterates over keys of an object, while for ... of iterates over values of an iterable object which. for ... in behavior can always be achieved with `Object.keys()` which IMHO is more clear

```suggestion
    "no-restricted-syntax": [
      "error",
      {
        selector: "ForInStatement",
        message: "for ... in disallowed, use for ... of instead",
      },
    ],

```

.eslintrc
```diff
@@ -0,0 +1,58 @@
...
+    "array-callback-return": "off",
+    "implicit-arrow-linebreak": "off",
+    "@typescript-eslint/no-non-null-asserted-optional-chain": "off",
+    "@typescript-eslint/no-non-null-assertion": "off",
+    "no-restricted-imports": "off",
+    "no-restricted-exports": "off",
+    "@typescript-eslint/ban-ts-comment": "off",
+    // TO FIX:
+    "import/no-cycle": "off", // pathpidia issue
+    "import/prefer-default-export": "off"
```
witoszekdev, line 46: Those 2 rules are really great when working with requests and responses (it might not be a concern for app-sdk but it saved me multiple times in saleor-app-checkout 😅) - basically if you have a function that returns a promise you *must* await it or use `void` to explicitly mark the promise as ignored. You also cannot use promise in `if` statement + it covers other pitfalls

```ts
const myValue = getPromise(); // oh no I forgot to await
if (myValue) {} // this will always be true
```


```suggestion
    "import/prefer-default-export": "off",
    "@typescript-eslint/no-misused-promises": ["error"],
    "@typescript-eslint/no-floating-promises": ["error"]

```

witoszekdev (commented):
I've proposed 3 rules that I found to be usefull in saleor-checkout ;)

tsconfig.json
```diff
@@ -25,7 +27,7 @@
     // "moduleDetection": "auto",                        /* Control what method is used to detect module-format JS files. */
 
     /* Modules */
-    "module": "commonjs",                                /* Specify what module code is generated. */
```
lkostrowski, line 28: Maybe we should remove these comments from `tsconfig`? It's autogenerated but once you set this up, it will probably stay not touched forever.

lkostrowski (approved):

tsconfig.json
krzysztofwolski: Personally, I'm ok with autogenerated commented JSON descriptions - Found them helpful many times.

PR #16, opened 2022-08-03
JWT validation middleware

### LINKED ISSUE #10: Missing support for validating JWT token passed by iframe
Saleor app template has the function for verifying the JWT token(https://github.com/saleor/saleor-app-template/blob/main/lib/middlewares.ts#L37). 
This is something directly related to the communication App <-> Saleor. Maybe it would be worth having implementation for this in app-sdk.

@magul @zaiste What do you think?

krzysztofwolski:
Closes #10
Migrate JWT security middleware to the SDK repo.

App template usage example:
https://github.com/saleor/saleor-app-template/pull/66

### DIFF
--- a/src/headers.ts
+++ b/src/headers.ts
@@ -1,0 +1,13 @@
+import {
+  SALEOR_AUTHORIZATION_BEARER_HEADER,
+  SALEOR_DOMAIN_HEADER,
+  SALEOR_EVENT_HEADER,
+  SALEOR_SIGNATURE_HEADER,
+} from "./const";
+
+export const getSaleorHeaders = (headers: { [name: string]: any }) => ({
+  domain: headers[SALEOR_DOMAIN_HEADER],
+  authorizationBearer: headers[SALEOR_AUTHORIZATION_BEARER_HEADER],
+  signature: headers[SALEOR_SIGNATURE_HEADER],
+  event: headers[SALEOR_EVENT_HEADER],
+});
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -1,6 +1,8 @@
 import crypto from "crypto";
 import * as jose from "jose";
-import type { Middleware } from "retes";
+import jwt, { JwtPayload } from "jsonwebtoken";
+import jwks, { CertSigningKey, RsaSigningKey } from "jwks-rsa";
+import type { Middleware, Request } from "retes";
 import { Response } from "retes/response";
 
 import { SALEOR_DOMAIN_HEADER, SALEOR_EVENT_HEADER } from "./const";
@@ -33,7 +35,6 @@ export const withSaleorEventMatch =
   (handler) =>
   async (request) => {
     const receivedEvent = request.headers[SALEOR_EVENT_HEADER];
-
     if (receivedEvent !== expectedEvent) {
       return Response.BadRequest({
         success: false,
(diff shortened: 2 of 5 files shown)


Cloud11PL (approved):

src/middleware.ts
```diff
@@ -106,3 +107,84 @@ export const withWebhookSignatureVerified =
...
+
+export interface DashboardTokenPayload extends JwtPayload {
+  app: string;
+}
+
+export const withJWTVerified =
+  (getAppId: (request: Request) => Promise<string | undefined>): Middleware =>
+  (handler) =>
+  async (request) => {
+    const { [SALEOR_DOMAIN_HEADER]: saleorDomain, "authorization-bearer": token } = request.headers;
```
lkostrowski, line 119: Can we move `authorization-bearer` to some const? I think it's reused and this is not type safe unfortunately

src/middleware.ts
```diff
@@ -106,3 +107,84 @@ export const withWebhookSignatureVerified =
...
+
+    if (token === undefined) {
+      return Response.BadRequest({
+        success: false,
+        message: "Missing token.",
```
lkostrowski, line 124: Maybe we can be explicit on errors, especially in SDK.
How about `Missing token. Check "authorization-bearer" header`

src/middleware.ts
```diff
@@ -106,3 +107,84 @@ export const withWebhookSignatureVerified =
...
+      return Response.BadRequest({
+        success: false,
+        message: "Invalid token.",
+      });
+    }
+
+    if ((tokenClaims as DashboardTokenPayload).iss !== saleorDomain) {
+      return Response.BadRequest({
+        success: false,
+        message: "Invalid token.",
```
lkostrowski, line 149: Also being explicit. We can say that token domain doesn't match or something like this?

src/middleware.ts
```diff
@@ -106,3 +107,84 @@ export const withWebhookSignatureVerified =
...
+    }
+
+    if (tokenClaims === null) {
+      return Response.BadRequest({
+        success: false,
+        message: "Invalid token.",
+      });
+    }
+
+    if ((tokenClaims as DashboardTokenPayload).iss !== saleorDomain) {
```
lkostrowski, line 146: Let's move this assertion, so we don't duplicate it

```
const dashboardTokenPayloadClaims = tokenClaims as DashboardTokenPayload;
```

lkostrowski (commented):
I'd add some explicit messages. In the future, I'd introduce some debug package (`DEBUG=*`) that will improve DX on local

src/middleware.ts
```diff
@@ -106,3 +107,84 @@ export const withWebhookSignatureVerified =
...
+    } catch (error) {
+      console.error("Error during getting the app ID.");
+      console.error(error);
+      return Response.BadRequest({
+        success: false,
+        message: "Error during token invalidation - could not obtain the app ID.",
+      });
+    }
+
+    if (!appId || (tokenClaims as DashboardTokenPayload).app !== appId) {
```
magul, line 165: if `!appId` - how does it mean, that token is invalid? shouldn't that be considered a 500 Internal Server Error, if our backend asking for `appId` using stored `auth_token` cannot resolve that to something defined?

src/middleware.ts
```diff
@@ -106,3 +107,84 @@ export const withWebhookSignatureVerified =
...
+      return Response.BadRequest({
+        success: false,
+        message: "Invalid token.",
+      });
+    }
+
+    const jwksClient = jwks({
+      jwksUri: `https://${saleorDomain}/.well-known/jwks.json`,
```
magul, line 173: this should use https://github.com/saleor/saleor-app-sdk/blob/main/src/urls.ts#L1 for determining whether to use `http` or `https`. Otherwise you will get some obscure error when trying to debug/develop using locally deployed sailor under http://localhost:8000 (I know, that in template it's that that way, I had just at the top of my personal backlog to take care of that, when moving to SDK).

src/middleware.ts
```diff
@@ -106,3 +107,84 @@ export const withWebhookSignatureVerified =
 
     return handler(request);
   };
+
+export interface DashboardTokenPayload extends JwtPayload {
+  app: string;
+}
+
+export const withJWTVerified =
+  (getAppId: (request: Request) => Promise<string | undefined>): Middleware =>
```
magul, line 116: As I understand, that's because you decided to not include GraphQL query for even that simple purpose in SDK, just abstracting that away and requiring template to inject that logic here.

The same way we did that with @witoszekdev in #8 ?

witoszekdev: yup, we didn't want to include codegen in that repository in order to avoid the need of releases whenever Saleor Core API changes.

src/middleware.ts
krzysztofwolski: You are right, such error would be much better

src/middleware.ts
krzysztofwolski: Additional bonus - the app's author can implement caching of the app id in provided function. Avoiding additional call to the GQL API would improve performance.

src/headers.ts
```diff
@@ -0,0 +1,13 @@
+import {
+  SALEOR_AUTHORIZATION_BEARER_HEADER,
+  SALEOR_DOMAIN_HEADER,
+  SALEOR_EVENT_HEADER,
+  SALEOR_SIGNATURE_HEADER,
+} from "./const";
+
+export const getSaleorHeaders = (headers: { [name: string]: any }) => ({
```
lkostrowski, line 8: Why not

```suggestion
export const getSaleorHeaders = (headers: { [name: string]: string }) => ({
```

lkostrowski (approved):

src/headers.ts
krzysztofwolski: I've followed the types from the Retes lib. I'll catch up with the authors what was the rationale behind any type and limit types in the separate PR

PR #17, opened 2022-08-04
Handle missing webhook signature header

### LINKED ISSUE #13: Missing saleor-signature header produces TypeError
It is probably a small thing, but when an App receives a webhook request without a `saleor-signature` header, the `TypeError` is raised (and 500 is returned):
`error - TypeError: Cannot read properties of undefined (reading 'split')`
It happens when `withWebhookSignatureVerified` is used.

krzysztofwolski:
fix #13

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -3,7 +3,7 @@ import * as jose from "jose";
 import type { Middleware, Request } from "retes";
 import { Response } from "retes/response";
 
-import { SALEOR_AUTHORIZATION_BEARER_HEADER } from "./const";
+import { SALEOR_AUTHORIZATION_BEARER_HEADER, SALEOR_SIGNATURE_HEADER } from "./const";
 import { getSaleorHeaders } from "./headers";
 import { jwksUrl } from "./urls";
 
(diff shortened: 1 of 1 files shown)


src/middleware.ts
```diff
@@ -70,6 +70,13 @@ export const withWebhookSignatureVerified =
     const { [SALEOR_DOMAIN_HEADER]: saleorDomain, "saleor-signature": payloadSignature } =
       request.headers;
 
+    if (!payloadSignature) {
+      return Response.BadRequest({
+        success: false,
+        message: "Missing payload signature.",
```
magul, line 76: here you have `payload signature`, below we just refer that that entity as a `signature`. We should stick to one.

magul (approved):
It's going to work fine, but we need to remember there's subtle logic beneath here (that can be supported by designing where to put that middleware, but it's not explicitly shown in its body).

We have currently 3 cases with webhook payload signatures:
 * Saleor 3.4 and earlier - no `secretKey` means not header at all
 * Saleor 3.5 and later - no `secretKey` means usage of JWS
 * `secretKey` set means old, legacy signature.

In here, in middleware we only support 2 later ones (and that's fine, as in template webhooks are set using manifest, which was introduced after signing payload in webhooks wihout `sercreKey`), but the support for legacy Saleor with webhook without `secretKey` and therefore without signature at all can be supported by just removing `withWebhookSignatureVerified` middleware from it's API handler.

lkostrowski (approved):

PR #25, opened 2022-08-11
Include app-bridge package into app-sdk

lkostrowski:
https://github.com/saleor/saleor-app-sdk/discussions/21

This PR
- Merges app-bridge into `@saleor/app-sdk/app-bridge` and reexports all symbols with compatiblity
- Deprecates some names, that will be removed after migration
- Refactors some app-bridge code
- Adds unit tests

It has been tested with latest app-sdk and Dashboard (tag 3.6)


lkostrowski:
@Cloud11PL is it possible that you test it with some actual app - replacing `app-bridge` dependency with this local package on a branch, importing from `@saleor/app-sdk/app-bridge`? 

I'm not sure if build is configured properly, but I don't have enough knowledge to test it correctly at the moment

Cloud11PL:
@lkostrowski Well, I can import but...: 
```
(property) ProviderProps<IAppContext>.value: IAppContext
Type '{ app: AppBridge; } | { app: undefined; }' is not assignable to type 'IAppContext'.
  Type '{ app: AppBridge; }' is not assignable to type 'IAppContext'.
    Types of property 'app' are incompatible.
      Property 'prototype' is missing in type 'AppBridge' but required in type 'typeof AppBridge'.ts(2322)
index.d.ts(328, 9): The expected type comes from property 'value' which is declared here on type 'IntrinsicAttributes & ProviderProps<IAppContext>'
```

lkostrowski:
@Cloud11PL thanks for your test. I fixed export and now it looks ok. 

I indeed deprecated createApp and few other symbols, because AppBridge now is a "module" inside SDK, so the soon we update names, the better

Somehow `this.refererOrigin = new URL(document.referrer).origin;` also was fixed for me.

>I think it would be great if we add at some point some more info for errors such as App token missing. like "Check if you're running latest Dashboard version" etc.

Definitely, we should keep updating such stuff all the time

Cloud11PL:
@lkostrowski Importing was fixed but when I launch the app outside of dashboard I get the following error:
image

Cloud11PL:
@lkostrowski 
image

lkostrowski:
@Cloud11PL I try it on Dashboard 3.6 and app-template and the same error occurs without this PR changes. Any ideas?

krzysztofwolski:
@lkostrowski Tried to work with the PR locally yesterday, but no luck - having the same issues as @Cloud11PL with cloud instance (dashboard v3.5.2 core v3.5.12)

- the "postMessage" error doesn't seem to be the cause here since it was occurring with the old SDK too 
- `listenOnMessages` method is called, but the event listener is not triggered even once, don't know why (without that token is not updated)

Code itself LGTM

Cloud11PL:
@krzysztofwolski @lkostrowski I found a fix, I'll be commiting soon

lkostrowski:
@Cloud11PL @krzysztofwolski problems are solved

1. Domain was not set and this is fixed and react works in production mode
2. React 18 causes problems - fixed here https://github.com/saleor/saleor-app-template/pull/78

krzysztofwolski:
Issues resolved! 🎉

PR #28, opened 2022-08-25
Introduce APL interface

krzysztofwolski:
Introduce a common interface for auth data.

### DIFF
--- a/docs/apl.md
+++ b/docs/apl.md
@@ -1,0 +1,68 @@
+# APL - Auth Persistance Layer
+
+APL is an interface for managing auth data of registered Apps. Implementing it does not restrict you from choosing the right technology for your project (for example Redis, Postgres, S3, etc.), and provides a common set of operations which can be used by your application and functions provided by App SDK.
+
+## Available methods
+
+- `get: (domain: string) => Promise<AuthData | undefined>` - If the entry for given domain exists, returns AuthData (`{ domain: string, token: string }`) object.
+
+- `set: (authData: AuthData) => Promise<void>` - Save auth data.
+
+- `delete: (domain: string) => Promise<void>` - Remove auth data fot the given domain.
+
+- `getAll: () => Promise<AuthData[]>` - Returns all auth data available.
+
+## Example implementation
+
+Let's create an APL, which uses redis for data storage:
+
+```ts
+import { createClient } from "redis";
+import { APL, AuthData } from "@saleor/app-sdk/types/apl";
+
+const client = createClient();
+await client.connect();
+
+const redisAPL: APL = {
+  get: async (domain: string) => {
+    const token = await client.get(domain);
+    if (token) {
+      return { token, domain };
+    }
+    return;
+  },
+  set: async (authData: AuthData) => {
+    await client.set(authData.domain, authData.token);
+  },
+  delete: async (domain: string) => {
+    await client.del(domain);
+  },
+  getAll: async () => {
+    throw new Exception("Not implemented.");
+  },
+};
+```
+
+You'll be able to use it directly:
+
+```ts
+import { redisAPL } from "./apl";
+
+const createTestData = async () => {
+  await redisAPL.set({ domain: "example.com ", token: "test-token" });
+};
+```
+
+And with middleware from the SDK:
+
+```ts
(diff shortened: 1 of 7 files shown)


docs/apl.md
```diff
@@ -0,0 +1,68 @@
...
+    }
+    return;
+  },
+  set: async (authData: AuthData) => {
+    await client.set(authData.domain, authData.token);
+  },
+  delete: async (domain: string) => {
+    await client.del(domain);
+  },
+  list: async () => {
```
lkostrowski, line 40: getAll

package.json
```diff
@@ -53,6 +53,11 @@
       "import": "./const.mjs",
       "require": "./const.js"
     },
+    "./types": {
+      "types": "./types.d.ts",
+      "import": "./types.mjs",
+      "require": "./types.js"
```
lkostrowski, line 59: Can we rather scope by packages instead by file types? It's not cohesive.

We can have `import {APL, ...} from '@saleor/app-sdk/APL'` and avoid mixing thing around the project.

The same I did in #25

src/manifest.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export interface WebhookManifest {
+  name: string;
+  events: string[];
+  query: string;
+  targetUrl: string;
+  isActive: boolean;
+}
+
+export interface Manifest {
```
lkostrowski, line 17: `AppManifest`? If we scope `ExtensionManifest` and `WebhookManifest`, maybe this also should be very explicit

lkostrowski: Also - maybe not in this PR - but I'd like to have Manifest strongly validated

src/middleware.ts
```diff
@@ -187,3 +188,19 @@ export const withJWTVerified =
 
     return handler(request);
   };
+
+export const withExistingRegistration =
```
lkostrowski, line 192: 1. Can we work on the naming better? I'm not sure if with is clear
2. Can we have some folder `middleware/` and there 1 file per middleware + test + index that exports them?

src/middleware.ts
```diff
@@ -187,3 +188,19 @@ export const withJWTVerified =
...
+  (apl: APL): Middleware =>
```
lkostrowski, line 193: can we decide that we will use `{apl: APL}` object for named parameters? It's not only better from DX perspective (I know exlicitly what I pass there), but also will protect us from breaking changes and multiple arguments in the future (we can just extend object)

src/types/apl.ts
```diff
@@ -0,0 +1,11 @@
+export interface AuthData {
+  domain: string;
+  token: string;
+}
+
+export interface APL {
```
lkostrowski, line 6: This also I'd move to `src/APL/apl.ts` and near it, we can have `src/APL/local-env-apl.ts` and other implementations

lkostrowski (changes requested):

src/manifest.ts
krzysztofwolski: Was not intended to be added here. Removing from the scope.

src/middleware.ts
krzysztofwolski: 1. Changed to a more explicit name, is it better?
2. Okay! Can it be a separate PR imminently after this one? In this one, I've followed the existing structure and would like to pollute APL changes with middleware refactor

krzysztofwolski: Issue for middleware [#29](https://github.com/saleor/saleor-app-sdk/issues/29)

src/middleware.test.ts
```diff
@@ -29,4 +32,54 @@ describe("middleware.test.ts", () => {
...
+      const mockRequest = {
+        context: {},
+        headers: {
+          host: "my-saleor-env.saleor.cloud",
+          "x-forwarded-proto": "https",
+          [SALEOR_DOMAIN_HEADER]: "not-registered.example.com",
+        },
+      } as unknown as Request;
+
+      await withRegisteredSaleorDomainHeader({ apl: mockAPL })(mockHandlerFn)(mockRequest);
```
lkostrowski, line 80: I think you can actually assert if error response was returned - this is client facing stuff

lkostrowski (approved):

PR #30, opened 2022-08-25
Add types for app manifest

witoszekdev:
I want to merge this because it adds types that can be used to make assertions for app's manifest.

For example:

```ts
// src/api/manifest.ts
import { AppManifest } from "@saleor/app-sdk/types";

const handler = (req, res) => {
  const manifest = {
      id: "saleor.checkout.app",
      version: version,
      name: appName,
      about: "Saleor checkout app to quickly configure and customize checkout in your store.",
      permissions: ["HANDLE_PAYMENTS", "HANDLE_CHECKOUTS", "MANAGE_ORDERS", "MANAGE_CHECKOUTS"],
      appUrl: baseURL,
      dataPrivacyUrl: `${baseURL}/data-privacy`,
      supportUrl: `${baseURL}/support`,
      tokenTargetUrl: `${baseURL}/api/register`,
    } as AppManifest; // <- type assertion goes here
  }

  res.json(manifest);
}

export default handler;
```

If a developer would forget about a required field (for example `id`), they would be notified immediately by TypeScript compiler.

It also makes the discovery of each field and it's documentaiton much easier (just hover a field to see the documentation).

Documentation for the fields was taken from [docs.saleor.io](https://docs.saleor.io/docs/3.x/developer/extending/apps/manifest)

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,0 +1,112 @@
+export type AppExtensionTarget = "POPUP" | "APP_PAGE";
+
+export type AppExtensionMount =
+  | "PRODUCT_DETAILS_MORE_ACTIONS"
+  | "PRODUCT_OVERVIEW_CREATE"
+  | "PRODUCT_OVERVIEW_MORE_ACTIONS"
+  | "NAVIGATION_CATALOG"
+  | "NAVIGATION_ORDERS"
+  | "NAVIGATION_CUSTOMERS"
+  | "NAVIGATION_DISCOUNTS"
+  | "NAVIGATION_TRANSLATIONS"
+  | "NAVIGATION_PAGES"
+  | "ORDER_DETAILS_MORE_ACTIONS"
+  | "ORDER_OVERVIEW_CREATE"
+  | "ORDER_OVERVIEW_MORE_ACTIONS";
+
+export type AppPermission =
+  | "HANDLE_PAYMENTS"
+  | "HANDLE_CHECKOUTS"
+  | "MANAGE_APPS"
+  | "MANAGE_CHECKOUTS"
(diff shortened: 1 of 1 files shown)


Cloud11PL (approved):

lkostrowski:
Related issue 
https://github.com/saleor/saleor-app-sdk/issues/14

krzysztofwolski (approved):
Thanks for the PR! LGTM

src/types.ts
```diff
@@ -0,0 +1,112 @@
...
+  events?: WebhookEvent[];
+  query: string;
+  targetUrl: string;
+  isActive?: boolean;
+}
+
+export interface AppManifest {
+  /** ID of the application used internally by Saleor */
+  id: string;
+  version: string;
```
lkostrowski, line 69: @krzysztofwolski maybe we don't support it yet but we should probably do `${number}.${number}.${number}`? I don't know if Semver is a way to go, but we need standardized versions to keep compatbility

krzysztofwolski:
@lkostrowski The change can be merged now and the class you mentioned will be submitted in the separate PR.

lkostrowski (approved):
Can be merged 👍

src/types.ts
krzysztofwolski: I would go with the semver. If used right, it gives good information if app upgrade is safe. Version string validation would by handled by the class

lkostrowski: Good. Runtime validation yes, but we can also add it in typescript recently, so I'd also add it
https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html

Issue #31, opened 2022-08-25
Test JWT middleware 


krzysztofwolski:
To test JWT middleware, we'll need PollyJS for mocking external calls

PR #32, opened 2022-08-25
Refactor and test middleware

### LINKED ISSUE #29: Update file structure for the middlewares
- Follow the file structure introduced in #25 
- Split middleware into separate files

krzysztofwolski:
Closes #29 

- split middleware into separate files
- add tests to existing middleware
- JWT middleware is not tested, and require more testing infrastructure (polly) to mock external requests ( issue #31 )


src/middleware/with-base-url.ts
```diff
@@ -0,0 +1,9 @@
+import { Middleware } from "retes";
+
+export const withBaseURL: Middleware = (handler) => async (request) => {
+  const { host, "x-forwarded-proto": protocol = "http" } = request.headers;
```
lkostrowski, line 4: I know its not a scope of refactor, but why http, not https?

lkostrowski (approved):

src/middleware/with-base-url.ts
krzysztofwolski: IIRC it's assuming http because of Docker routing.

PR #35, opened 2022-08-29
Introduce new APLs: fileAPL and vercelAPL

### LINKED ISSUE #33: Implement environment variables APL
- Migrate code from template https://github.com/saleor/saleor-app-template/blob/main/lib/environment.ts
- Prepare documentation describing when to use this APL
- Add jsDoc to the implementation

krzysztofwolski:
Closes #33

The old implementation from the app-template had a single implementation for both use cases: local env and vercel. Decided to split it into separate APLs, which resulted in a simpler, one-purpose code.

Now the choice of storage will be on the App developer, for example:
```typescript
import { fileAPL, vercelAPL } from "@saleor/app-sdk/APL";

// For local development store auth data in the `.envfile`.
// For app deployment on Vercel with Saleor CLI, use vercelAPL.
export const apl = process.env.VERCEL === "1" ? vercelAPL : fileAPL;
```
( https://github.com/saleor/saleor-app-template/pull/79 )


krzysztofwolski:
Usage example: https://github.com/saleor/saleor-app-template/pull/79

src/APL/fileAPL.ts
```diff
@@ -0,0 +1,85 @@
+import { promises as fsPromises } from "fs";
+
+import { APL, AuthData } from "./apl";
+
+const ENV_FILE_NAME = ".envfile";
```
lkostrowski, line 5: I think `.env` is a common standard, but I'd use https://www.npmjs.com/package/dotenv to manage them

src/APL/fileAPL.ts
```diff
@@ -0,0 +1,85 @@
...
+export const fileAPL: APL = {
+  get: async (domain) => {
+    const authData = await loadDataFromFile();
+    if (domain === authData?.domain) {
+      return authData;
+    }
+    return undefined;
+  },
+  set: async (authData: AuthData) => {
+    console.debug("Setting up variable", authData);
```
lkostrowski, line 66: can you use introduced debug package?

src/APL/fileAPL.ts
```diff
@@ -0,0 +1,85 @@
...
+const DOMAIN_VARIABLE_NAME = "SALEOR_DOMAIN";
+
+const saveDataToFile = async (authData?: AuthData) => {
+  let currentEnvVars;
+  try {
+    await fsPromises.access(ENV_FILE_NAME);
+    currentEnvVars = JSON.parse(await fsPromises.readFile(ENV_FILE_NAME, "utf-8"));
+  } catch {
+    currentEnvVars = {};
+  }
```
lkostrowski, line 16: missing error handling for file access, also - better use dotenv package?

src/APL/fileAPL.ts
lkostrowski: BTW this probably should be configurable in APL instance. 

I'd do `fileAPL = new FileApl(config: {envFile: string})` etc

lkostrowski (commented):
Partial review before my day off :)

src/APL/fileAPL.ts
krzysztofwolski: I'll defend the current implementation:
- it's not in the `.env` (dotenv) format - we are serializing a JSON file, so the syntax is different. To have the same syntax, I would need to use an additional package... 
- ...and `dotenv`  has support *only* for reading files. Dumping variables to the existing file would require our own serializer
- even if we had a serializer, it would mean our code would mess with `.env` file which could be created by the developer. And overriding would be tricky - we would not like to change other values (or override comments in it)
- Adding an additional package doesn't seem to be worth it in this case, where we only load/store json for this dedicated purpose

Regarding configuration - good idea

lkostrowski:
@krzysztofwolski can you please extend docs and add some section about usage - especially that app will locally use local env, but on vercel use its implementation?

src/APL/fileAPL.ts
```diff
@@ -0,0 +1,87 @@
+import debugPkg from "debug";
+import { promises as fsPromises } from "fs";
+
+import { APL, AuthData } from "./apl";
+
+const debug = debugPkg.debug("FileAPL");
```
lkostrowski, line 6: @zaiste had a good idea to nest namespaces - what do you think about

Level 1 - namespace `@saleor/app-sdk` or `app-sdk`
Level 2 - namespace for module `FileAPL`

So final log will be `app-sdk:FileAPL: Something Happened`

I link this idea because users can filter either entire package or selected modules

src/APL/fileAPL.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export const loadDataFromFile = async (fileName: string): Promise<AuthData | undefined> => {
```
lkostrowski, line 8: For me it looks like a private method that is exported just to be tested? If so, I'd not to - private methods should be tested indirectly through public methods.

(this is problem not using classes, functions tend to leak due to lack of private keywords)

src/APL/fileAPL.ts
```diff
@@ -0,0 +1,87 @@
...
+};
+
+export const saveDataToFile = async (fileName: string, authData?: AuthData) => {
+  debug(`Save auth data to the ${fileName} file`);
+  const newData = authData ? JSON.stringify(authData) : "{}";
+  try {
+    await fsPromises.writeFile(fileName, newData);
+  } catch (err) {
+    debug(`Could not save auth data to the ${fileName} file`, err);
+  }
```
lkostrowski, line 32: Error is hidden silently here. App can't work if the file is not writeable. So as a user I'd expect it to fail fast - throw as soon as possible, telling me that I must ensure permissions are correct. Otherwise there will be hidden bugs why auth doesn't work

src/APL/fileAPL.ts
```diff
@@ -0,0 +1,87 @@
...
+ *   - it's not recommended for production use - redeployment of the application will override
+ *     existing values, or data persistence will not be guaranteed at all depending on chosen
+ *     hosting solution
+ *
+ */
+export class FileAPL implements APL {
+  private fileName: string;
+
+  constructor(config: FileAPLConfig = {}) {
+    this.fileName = config?.fileName || ".auth-data.json";
```
lkostrowski, line 55: How about `.saleor-app-auth.json`? 

1. Having maaany files you are sure this is Saleor, there can be many other "auths"
2. `Data` is forbidden keyword meaning nothing in programming ;)

src/APL/fileAPL.ts
```diff
@@ -0,0 +1,87 @@
...
+ * The APL store auth data in the json file.
+ *
+ * Before using this APL, please take in consideration:
+ *   - only stores single auth data entry (setting up a new one will overwrite previous values)
+ *   - it's not recommended for production use - redeployment of the application will override
+ *     existing values, or data persistence will not be guaranteed at all depending on chosen
+ *     hosting solution
+ *
+ */
+export class FileAPL implements APL {
```
lkostrowski, line 51: If you created class, why functions above are not private members of it?

src/APL/vercelAPL.ts
```diff
@@ -0,0 +1,116 @@
...
+import fetch from "node-fetch";
+
+import { APL, AuthData } from "./apl";
+
+const debug = debugPkg.debug("VercelAPL");
+
+export const TOKEN_VARIABLE_NAME = "SALEOR_AUTH_TOKEN";
+export const DOMAIN_VARIABLE_NAME = "SALEOR_DOMAIN";
+export const SALEOR_REGISTER_APP_URL = "SALEOR_REGISTER_APP_URL";
+export const SALEOR_DEPLOYMENT_TOKEN = "SALEOR_DEPLOYMENT_TOKEN";
```
lkostrowski, line 12: What do you think about some namespace object, like `VercelAplConfig` or something like this, so you will have:

```
import {VercelAplConfig} ...

VercelAplConfig.TOKEN_VARIABLE_NAME
```

This way we:
1. Are very explicit this is Vercel, not e.g. Netlify or something else (import paths are not good enough, because IDE is autogenerating them) - SDK will grow so this way we can mimic scoping/namespaces from other languages
2. Make it cohesive - put all related variables into single space where each of them can be found
3. Decrease number of exported symbols which is a problem scaling up

src/APL/vercelAPL.ts
```diff
@@ -0,0 +1,116 @@
...
+
+const envAuthData = (): AuthData | undefined => {
```
lkostrowski, line 14: Can we name functions with verbs so they are explicit? In JS `envAuthData` can be a function but also some static object, with `getEnvAuthData` (or better `getEnvAuth`) its clear that this is a function

src/APL/vercelAPL.ts
```diff
@@ -0,0 +1,116 @@
...
+    domain,
+  };
+};
+
+const saveDataToVercel = async (
+  registerAppURL: string,
+  deploymentToken: string,
+  authData?: AuthData
+) => {
+  debug("Saving data to Vercel");
```
lkostrowski, line 31: I'm not sure if this is a good debug message. Shouldn't it point to specific actions, like `saveDataToVercel() called with: ...`?

src/APL/vercelAPL.ts
```diff
@@ -0,0 +1,116 @@
...
+      }),
+    });
+  } catch (error) {
+    debug("Error during saving the data:", error);
+  }
+};
+
+export type VercelAPLConfig = {
+  registerAppURL?: string;
+  deploymentToken?: string;
```
lkostrowski, line 52: Can you add JSDoc to these too? It will explain users, can link to docs, will be available in IDE

src/APL/vercelAPL.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  constructor(config?: VercelAPLConfig) {
+    const registerAppURL = config?.registerAppURL || process.env.SALEOR_REGISTER_APP_URL;
+    if (!registerAppURL) {
+      debug("Missing registerAppURL");
+      throw new Error("Misconfiguration: please provide registerAppUrl");
+    }
+    const deploymentToken = config?.deploymentToken || process.env.SALEOR_DEPLOYMENT_TOKEN;
+    if (!deploymentToken) {
+      debug("Missing deploymentToken");
```
lkostrowski, line 81: I think debug message here is redundant because Error will always follow with more detailed message

src/APL/vercelAPL.test.ts
```diff
@@ -0,0 +1,145 @@
...
+        expect(() => new VercelAPL()).toThrow();
+
+        process.env[SALEOR_REGISTER_APP_URL] = "http://example.com";
+        delete process.env[SALEOR_DEPLOYMENT_TOKEN];
+
+        expect(() => new VercelAPL()).toThrow();
+      });
+    });
+
+    it("Constructor with config values", async () => {
```
lkostrowski, line 33: Test name can be improved - something like "Constructs VercelAPL instance when deploymentToken and registerAppURL provided"

src/APL/vercelAPL.test.ts
```diff
@@ -0,0 +1,145 @@
...
+      expect(
+        () =>
+          new VercelAPL({
+            deploymentToken: "token",
+            registerAppURL: "http://example.com",
```
lkostrowski, line 38: not pushing, but maybe we can easily validate if URL is valid syntax?

src/APL/vercelAPL.test.ts
```diff
@@ -0,0 +1,145 @@
...
+          })
+      ).not.toThrow();
+    });
+
+    it("Constructor with config values from environment variables", async () => {
```
lkostrowski, line 43: Maybe there can be another test that constructor options will take precedence over env variables?

src/APL/vercelAPL.test.ts
```diff
@@ -0,0 +1,145 @@
...
+      expect(() => new VercelAPL()).not.toThrow();
+    });
+
+    describe("get", () => {
+      describe("Read existing auth data from env", () => {
+        const initialEnv = { ...process.env };
+
+        afterEach(() => {
+          process.env = { ...initialEnv };
+        });
```
lkostrowski, line 56: Maybe you can lift it to top of root describe?

lkostrowski (changes requested):

src/APL/fileAPL.test.ts
```diff
@@ -0,0 +1,184 @@
...
+        expect(spyWriteFile).toBeCalledWith(
+          "test.json",
+          JSON.stringify({
+            domain: "example.com",
+            token: "example-token",
+          })
+        );
+      });
+
+      it("Save empty file when no auth data provided", async () => {
```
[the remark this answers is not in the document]
krzysztofwolski: That's expected - I use it for the data remove. Should be explained in the docstring 👍

src/APL/fileAPL.test.ts
```diff
@@ -0,0 +1,184 @@
...
+        vi.spyOn(fsPromises, "readFile").mockResolvedValue("{}");
+
+        const apl = new FileAPL();
+
+        expect(await apl.getAll()).toStrictEqual([]);
+      });
+    });
+  });
+
+  describe("FileAPL utils", () => {
```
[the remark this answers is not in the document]
krzysztofwolski: Theres a bit of logic in those functions - handle errors and make sure we return desired structure. Thought it would be better to test edge cases of this function alone. WDYT?

src/APL/fileAPL.test.ts
```diff
@@ -0,0 +1,184 @@
...
+        expect(spyWriteFile).toBeCalledWith("test.json", "{}");
+      });
+
+      it("Handle write file errors", async () => {
+        const spyWriteFile = vi.spyOn(fsPromises, "writeFile").mockImplementation(() => {
+          throw Error("Write error");
+        });
+        await saveDataToFile("test.json");
+        expect(spyWriteFile).toBeCalledWith("test.json", "{}");
+      });
```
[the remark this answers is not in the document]
krzysztofwolski: It handles errors silently :X.

How would you suggest to handle errors? APLs should return error value, or throw errors?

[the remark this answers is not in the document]
lkostrowski: IMO, SDK should throw. This is best for SDK user - they immediately knows what happened and they can fix (in this case - write access).

But the test should come from public method, I'd not test utils

src/APL/fileAPL.test.ts
```diff
@@ -0,0 +1,119 @@
+import { promises as fsPromises } from "fs";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import { FileAPL } from "./fileAPL";
+
+const stubAuthData = {
+  domain: "example.com",
+  token: "example-token",
+};
```
lkostrowski, line 9: protip (here it doesn't matter) - it's safer in tests to ensure immutability (eg method `getStubAuthData()`) that will create a fresh one on every test. 

technically, inside the test, we can't ensure that some tested code doesn't mutate/modify stub/mock and this object can leak to other test with mutated value

lkostrowski (approved):

src/APL/fileAPL.test.ts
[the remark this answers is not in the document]
lkostrowski: AFAIK its a bad pattern and need of testing private methods is also a sign of something wrong. Edge cases can be tested via public methods too. Private method will never be called directly by the client

lkostrowski (approved):

PR #36, opened 2022-08-29
Fix manifest type to use `asyncEvents` and `syncEvents`

witoszekdev:
This PR fixes manifest types to use `asyncEvents` and `syncEvents` instead of `events` proerty. See https://github.com/saleor/saleor-docs/pull/542

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -57,7 +57,8 @@ export interface AppExtension {
 
 export interface AppWebhook {
   name: string;
-  events?: WebhookEvent[];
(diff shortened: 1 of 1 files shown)


lkostrowski:
@krzysztofwolski @przlada can you check if this is right? :)

krzysztofwolski (approved):

krzysztofwolski:
@lkostrowski yup - it seems to be the same as in the inferWebhooks helper 👍

lkostrowski (approved):

PR #38, opened 2022-08-30
Remove `np` packge

lkostrowski:
Since `pnpm` is used for publishing, I removed `np` so it doesn't confuse us

Let's introduce `release-it` soon


lkostrowski:
@zaiste commit history cleaned up

krzysztofwolski (approved):

PR #39, opened 2022-08-30
Change "debug" package import to use default

lkostrowski:
This is rquired because app-template broke when using app-sdk with previous
named import

### DIFF
--- a/src/app-bridge/app-bridge.ts
+++ b/src/app-bridge/app-bridge.ts
@@ -1,4 +1,4 @@
-import { debug as createDebug } from "debug";
+import debugPkg from "debug";
 
 import { Actions } from "./actions";
 import { AppBridgeState, AppBridgeStateContainer } from "./app-bridge-state";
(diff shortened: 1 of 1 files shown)


krzysztofwolski (approved):

krzysztofwolski:
@lkostrowski could you add docs file with the description how to use debug? (local storage trick we used previously). It would be really helpful to promote this knowledge.

Docs are preferred, but opening an issue for docs will also be ok :)

zaiste:
Slight improvement: 

```tsx
import Debug from 'debug';
const debug = Debug('app:sdk'); // eslint-disable-line no-unused-vars
```

so that you can then enable/disable all for `app` with `DEBUG=app:*` or only for the `sdk` with `DEBUG=app:sdk:*`

WDYT?

lkostrowski:
@zaiste good idea, how about we scope by package name and then, local namespace, so:
`app-sdk:AppBridge` or `app-sdk:*`?

So we can filter by npm package/repo but also more fine grade

zaiste:
@lkostrowski yes, exactly, even better. Maybe the local namespace could the the file in the repo?

lkostrowski:
I don't think files are good because its implementation detail, we should be able to change them, they can also be quite ambiguous (util.ts? ;P)

I'd vote on some modules/class names, something like this.

michalina-graczyk:
@lkostrowski should it be checked by QA?

lkostrowski:
@michalina-graczyk no, this fixes error that failed on a build stage

PR #42, opened 2022-08-30
Install and configure release-it

lkostrowski:
Using setup from saleor dashboard.

### DIFF
--- a/.release-it.json
+++ b/.release-it.json
@@ -1,0 +1,16 @@
+{
+  "github": {
+    "release": true,
+    "draft": true,
+    "releaseName": "${version}"
+  },
+  "npm": {
+    "publish": false
+  },
+  "plugins": {
+    "@release-it/bumper": {
+      "in": "package.json",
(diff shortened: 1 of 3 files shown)


.release-it.json
```diff
@@ -0,0 +1,16 @@
+{
+  "github": {
+    "release": true,
+    "draft": true,
+    "releaseName": "${version}"
+  },
+  "npm": {
+    "publish": false
```
krzysztofwolski, line 8: Shouldn't it be set to true?

lkostrowski: Probably 🤔 In Dashboard it's like this. I think they open draft release and a Github action is publishing

In this release, I will set `publish: true` and we will improve

krzysztofwolski: Dashboard does not publish packages to npm (its a standalone app)

krzysztofwolski (approved):

lkostrowski:
@krzysztofwolski leaving to you for improvement

PR #45, opened 2022-09-01
Add debug messages to middlewares

lkostrowski:


src/middleware/with-webhook-signature-verified.ts
```diff
@@ -53,9 +66,15 @@ export const withWebhookSignatureVerified =
         new URL(getJwksUrl(saleorDomain))
       ) as jose.FlattenedVerifyGetKey;
 
+      debug("Created remote JWKS");
+
       try {
         await jose.flattenedVerify(jws, remoteJwks);
+
+        debug("JKWS verified");
```
krzysztofwolski, line 74:
```suggestion
        debug("JWKS verified");
```

src/middleware/with-webhook-signature-verified.ts
```diff
@@ -35,7 +44,11 @@ export const withWebhookSignatureVerified =
         .update(request.rawBody)
         .digest("hex");
 
+      debug("signature was calculated");
+
       if (calculatedSignature !== payloadSignature) {
+        debug("calculated signature doesnt match payload signature, will return Bad Request");
```
krzysztofwolski, line 50:
```suggestion
        debug("Calculated signature doesn't match payload signature, will return Bad Request");
```

src/middleware/with-webhook-signature-verified.ts
```diff
@@ -35,7 +44,11 @@ export const withWebhookSignatureVerified =
         .update(request.rawBody)
         .digest("hex");
 
+      debug("signature was calculated");
```
krzysztofwolski, line 47:
```suggestion
      debug("Signature was calculated");
```

src/middleware/with-webhook-signature-verified.ts
```diff
@@ -23,6 +30,8 @@ export const withWebhookSignatureVerified =
     const { domain: saleorDomain, signature: payloadSignature } = getSaleorHeaders(request.headers);
 
     if (!payloadSignature) {
+      debug("signature header was not found");
```
krzysztofwolski, line 33:
```suggestion
      debug("Signature header was not found");
```

src/middleware/with-webhook-signature-verified.ts
```diff
@@ -6,14 +6,21 @@ import { Response } from "retes/response";
...
 
 export const withWebhookSignatureVerified =
   (secretKey: string | undefined = undefined): Middleware =>
   (handler) =>
   async (request) => {
-    const ERROR_MESSAGE = "Webhook signature verification failed:";
+    debug("Middleware executing start");
 
     if (request.rawBody === undefined) {
+      debug("Request rawBody was not found, will return Internal Server Error");
```
krzysztofwolski, line 22:
```suggestion
      debug("Request rawBody was not found");
```

Don't we need to additionally describe what middleware will return - describing everything will get tedious very fast IMO.

src/middleware/with-jwt-verified.ts
```diff
@@ -5,71 +5,105 @@ import { Response } from "retes/response";
...
+      debug("tokenClaims couldnt be decoded from JWT, will respond with Bad Request");
+
       return Response.BadRequest({
         success: false,
         message: `${ERROR_MESSAGE} Could not decode authorization token.`,
       });
     }
 
     if (tokenClaims.iss !== domain) {
+      debug("token iss doesnt match domain, will response with Bad Request");
```
krzysztofwolski, line 52:
```suggestion
      debug("Token iss doesn't match domain, will response with Bad Request");
```

src/middleware/with-jwt-verified.ts
```diff
@@ -5,71 +5,105 @@ import { Response } from "retes/response";
...
       });
     }
 
+    debug("Middleware called with token starting with: \"%s\"", token.substring(0, 4));
+
     let tokenClaims: DashboardTokenPayload;
+
     try {
       tokenClaims = jose.decodeJwt(token as string) as DashboardTokenPayload;
+      debug("tokenClaims decoded from jwt");
```
krzysztofwolski, line 41:
```suggestion
      debug("TokenClaims decoded from JWT");
```

src/middleware/with-jwt-verified.ts
```diff
@@ -5,71 +5,105 @@ import { Response } from "retes/response";
...
     } catch (e) {
+      debug("tokenClaims couldnt be decoded from JWT, will respond with Bad Request");
```
krzysztofwolski, line 43:
```suggestion
      debug("TokenClaims couldn't be decoded from JWT, will respond with Bad Request");
```

src/middleware/with-auth-token-required.ts
```diff
@@ -1,9 +1,18 @@
 import { Middleware } from "retes";
 import { Response } from "retes/response";
 
+import { createMiddlewareDebug } from "./middleware-debug";
+
+const debug = createMiddlewareDebug("withAuthTokenRequired");
+
 export const withAuthTokenRequired: Middleware = (handler) => async (request) => {
+  debug("middleware called");
```
krzysztofwolski, line 9:
```suggestion
  debug("Middleware called");
```

src/middleware/with-auth-token-required.ts
```diff
@@ -1,9 +1,18 @@
...
+
   const authToken = request.params.auth_token;
+
   if (!authToken) {
+    debug("found missing authToken param");
```
krzysztofwolski, line 14:
```suggestion
    debug("Found missing authToken param");
```

krzysztofwolski (changes requested):

src/middleware/with-jwt-verified.ts
lkostrowski: Hmm, `tokenClaims` mimics variable name, `TokenClaims` is something that doesn't "exist". So maybe `Token claims` and detach from variable here?

src/middleware/with-webhook-signature-verified.ts
lkostrowski: I'm not sure. Response from middleware isn't traced anywhere right? Retrning handler is easy to debug because next messages will follow. Response is breaking the chain, and this error can be found in frontend. IMO it "closes" debug flow and you know that this response is over. 

Maybe we can implement a DebugMiddleware that will just optionally log responses?

krzysztofwolski: Like the idea of debugging middleware - it could log Request and Responses 👍

src/middleware/with-jwt-verified.ts
krzysztofwolski: Sounds good!

src/middleware/middleware-debug.ts
```diff
@@ -0,0 +1,26 @@
...
+    const debug = debugFactory(handler.name);
+
+    debug("Called with request %j", request);
+
+    const response = await handler(request);
+
+    debug("Responded with response %j", response);
+
+    return response;
+  };
```
lkostrowski, line 26: @krzysztofwolski I added this one, but I'd like not to connect it yet. Maybe it should be something that client is connecting explicitly? I'm worried that logging everything may leak data

src/middleware/with-jwt-verified.ts
lkostrowski: fixed

src/middleware/with-webhook-signature-verified.ts
lkostrowski: added experimental middleware

src/middleware/middleware-debug.ts
krzysztofwolski: Agree, that's a helper which I would like to use occasionally

PR #46, opened 2022-09-02
Fix lint warnings

krzysztofwolski:
- update types on headers helper function
- ignore type warnings in logging message
- use unknown instead of any for unknown events

### DIFF
--- a/src/app-bridge/app-bridge.ts
+++ b/src/app-bridge/app-bridge.ts
@@ -44,6 +44,7 @@ function eventStateReducer(state: AppBridgeState, event: Events) {
       /**
        * Event comes from API, so always assume it can be something not covered by TS
        */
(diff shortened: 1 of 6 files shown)


src/app-bridge/app-bridge.ts
```diff
@@ -44,6 +44,7 @@ function eventStateReducer(state: AppBridgeState, event: Events) {
       /**
        * Event comes from API, so always assume it can be something not covered by TS
        */
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
       console.warn(`Invalid event received: ${(event as any)?.type}`);
```
lkostrowski, line 48: Eslint is right here - it comes from API, so its type should be `unknown`

lkostrowski (approved):

PR #51, opened 2022-09-06
Add AppBridgeProvider and events factory

lkostrowski:


src/app-bridge/app-bridge-provider.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+   */
+  it.todo("Throws if not wrapped inside AppBridgeProvider");
+
+  it("Returned instance provided in Provider", () => {
+    const appBridge = new AppBridge({
+      targetDomain: "test-domain",
+    });
+
+    const { result } = renderHook(() => useAppBridge(), {
+      wrapper: (props: any) => <AppBridgeProvider {...props} appBridgeInstance={appBridge} />,
```
krzysztofwolski, line 59: Please add lint ignore or use different type

src/app-bridge/app-bridge-provider.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Ensures component re-renders each time even from Dashboard receives
+ */
+export const useReactiveAppBridge = () => {
+  const { app } = useAppBridge();
+  const [, setState] = useState(0);
+
+  const update = () => {
+    setState(Math.random());
```
krzysztofwolski, line 62: Why do we need that?

krzysztofwolski (changes requested):

src/app-bridge/app-bridge-provider.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+  it("App is defined when wrapped in AppBridgeProvider", async () => {
+    const { result } = renderHook(() => useAppBridge(), {
+      wrapper: AppBridgeProvider,
+    });
+
+    expect(result.current.app).toBeDefined();
+  });
+
+  /**
+   * TODO - How to write this test and avoid logging bloated error to console?
```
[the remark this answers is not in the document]
lkostrowski: Yes, good idea. Its strange that Vitest doesnt handle that, because its wrapped in expert error assertion

src/app-bridge/app-bridge-provider.tsx
lkostrowski: This is code moved from app-template. Its a dirty hack :) It rerenders each time any subscribed event is called by dashboard.

Maybe I can rewrite it in this PR. Its small change to pass getState() into useState and provided will synchronize everything 👍 🦆

src/app-bridge/app-bridge-provider.test.tsx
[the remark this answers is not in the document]
lkostrowski: Unfortunately, this is some renderHook thing, I can't catch this error in try/catch block :/ will leave it in todo state, maybe I will figure it out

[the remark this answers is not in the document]
lkostrowski: done

src/app-bridge/app-bridge-provider.tsx
lkostrowski: Done. Left only one hook, that is always reactive and it stores react state

krzysztofwolski (approved):

PR #52, opened 2022-09-06
Fix Vercel APL set method + add tests for it

krzysztofwolski:
- add missing tests for set method
- fix setting up variables

### DIFF
--- a/src/APL/vercel-apl.test.ts
+++ b/src/APL/vercel-apl.test.ts
@@ -1,6 +1,32 @@
-import { afterEach, describe, expect, it } from "vitest";
+import { rest } from "msw";
+import { setupServer } from "msw/node";
+import { afterAll, afterEach, beforeAll, describe, expect, it } from "vitest";
 
 import { VercelAPL, VercelAPLVariables } from "./vercel-apl";
+
+export const handlers = [
+  rest.post("https://registerService.example.com/internalError", async (req, res, ctx) =>
+    res(ctx.status(500))
+  ),
+  rest.post("https://registerService.example.com/", async (req, res, ctx) => {
+    const body = await req.json();
+    // Register service will return error when request does not provide token and envs
+    if (!body?.token || !body?.envs) {
+      return res(ctx.status(400));
+    }
+
+    // Check if envs follow the anticipated format
+    const envs = body.envs as Record<string, string | undefined>[];
+    for (const env of envs) {
+      if (typeof env.key === "undefined" || typeof env?.value === "undefined") {
+        return res(ctx.status(400));
+      }
+    }
+    return res(ctx.status(200));
+  }),
+];
+
+export const server = setupServer(...handlers);
 
 const aplConfig = {
   deploymentToken: "token",
@@ -15,8 +41,11 @@ const stubAuthData = {
 describe("APL", () => {
   const initialEnv = { ...process.env };
 
+  beforeAll(() => server.listen({ onUnhandledRequest: "error" }));
+  afterAll(() => server.close());
   afterEach(() => {
     process.env = { ...initialEnv };
+    server.resetHandlers();
   });
 
   describe("VercelAPL", () => {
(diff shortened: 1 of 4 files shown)


src/APL/vercel-apl.test.ts
```diff
@@ -1,7 +1,33 @@
-import { afterEach, describe, expect, it } from "vitest";
+import { rest } from "msw";
+import { setupServer } from "msw/node";
+import { afterAll, afterEach, beforeAll, describe, expect, it } from "vitest";
 
 import { VercelAPL, VercelAPLVariables } from "./vercel-apl";
 
+export const handlers = [
```
lkostrowski, line 7: shouldn't be exported

src/APL/vercel-apl.test.ts
```diff
@@ -1,7 +1,33 @@
...
+    }
+
+    // Check if envs follow the anticipated format
+    const envs = body.envs as Record<string, string | undefined>[];
+    for (const env of envs) {
+      if (typeof env.key === "undefined" || typeof env?.value === "undefined") {
+        return res(ctx.status(400));
+      }
+    }
+    return res(ctx.status(200));
```
lkostrowski, line 25: this body validation is an assumption, but has to be implemented in Appraptor cc @zaiste

src/APL/vercel-apl.test.ts
```diff
@@ -56,6 +85,26 @@ describe("APL", () => {
       expect(apl["registerAppURL"]).toBe("option");
     });
 
+    describe("set", () => {
+      it("Successful save of the auth data", async () => {
+        const apl = new VercelAPL({
+          registerAppURL: "https://registerService.example.com",
+          deploymentToken: "token",
+        });
+        expect(await apl.set({ domain: "example.com", token: "token" })).toBe(undefined);
```
lkostrowski, line 94: This test doesnt really check anything. It doesnt check if auth data was saved - it checks if promise returns undefined.

You can check if request was delivered to msw in proper format. Or rename test to something like "doesn't throw for XYZ input" - but then this test is not too helpful

src/APL/vercel-apl.ts
```diff
@@ -65,23 +65,31 @@ export class VercelAPL implements APL {
   }
 
   private async saveDataToVercel(authData?: AuthData) {
-    debug(`saveDataToVercel with: ${authData}`);
+    debug("saveDataToVercel with: %j", authData);
```
lkostrowski, line 68: Lets write debug messages that are explicit. If we track function called, it can be
`saveDataToVercel() called with "%j` - so its clear that debug says about function invocation. Its also most likely by clear that its beginning of the function

lkostrowski: Also shouldn't you obfuscate authdata? Maybe some substring(0,4) will work for token?

src/APL/vercel-apl.ts
```diff
@@ -65,23 +65,31 @@ export class VercelAPL implements APL {
...
+            { key: VercelAPLVariables.DOMAIN_VARIABLE_NAME, value: authData?.domain || "" },
+          ],
         }),
       });
     } catch (error) {
       debug("Error during saving the data:", error);
-      throw new Error(`VercelAPL was not able to save auth data${error}`);
+      throw new Error(`VercelAPL was not able to save auth data ${error}`);
     }
+    if (response.status !== 200) {
```
lkostrowski, line 86: Its risky, technically Appraptor can return 202 and 204 (which will be valid here).

Safely to check if >=400 || <200

src/APL/vercel-apl.ts
```diff
@@ -65,23 +65,31 @@ export class VercelAPL implements APL {
...
+      debug("Non 200 response code. Register service responded with %j", response);
+      throw new Error(
+        `Vercel APL was not able to save auth data, register service responded with the code ${response.status}`
+      );
+    }
+    debug("Register service called successfully");
```
lkostrowski, line 92: ~ called~ -> responded? Call was always successful

lkostrowski (changes requested):

src/APL/vercel-apl.test.ts
krzysztofwolski: That's an assumption but wanted to cover the issue I fix with this PR.

lkostrowski: Good

lkostrowski (approved):

PR #56, opened 2022-09-13
Set pre-release prefix to be RC

lkostrowski:


lkostrowski:
Lets first figure out entire flow, including npm tags, prerelease and pnpm forcing

PR #59, opened 2022-09-14
Add doc with proposal of MetadataManager api

lkostrowski:
This is a proposal for apps/envs cloning, that will move responsibility to apps themselves

### DIFF
--- a/docs/shared-metadata.md
+++ b/docs/shared-metadata.md
@@ -1,0 +1,44 @@
+# Cross-env metadata
+
+To achieve Saleor (core) cloning functionalities (e.g. clone staging to production),
+apps must be able to maintain their state (config, keys).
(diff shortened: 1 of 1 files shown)


lkostrowski:
@krzysztofwolski based on discussion this probably will use something like this (in metadata):

```
key: {
[domain: string]: {
 value: string,
 ....
}
}
```

krzysztofwolski:
@lkostrowski: Prepared PR with manager proposal: https://github.com/saleor/saleor-app-template/pull/89

Wanted to avoid serializing values back and forth and proposed to use convention for keys:
    - for env specific `key__domain.com`
    - for rest `key`

IMO with such an approach, we can have a more straightforward implementation.

PR #61, opened 2022-09-15
Add Handler factories

lkostrowski:
This PR contributes to lean app templates, allowing sleek and minimal creation of handlers in Next

In the future, we will transform SDK to monorepo with SDK core + separate React, Next, Remix etc

### DIFF
--- a/docs/api-handlers.md
+++ b/docs/api-handlers.md
@@ -11,8 +11,6 @@ Saleor requires 2 endpoints to be available for a standalone app:
 - Manifest endpoint - Returns JSON object with app properties, like its name or permissions. [Read more](https://docs.saleor.io/docs/3.x/developer/extending/apps/manifest)
 - Register endpoint - During the installation process, Saleor sends `POST` request with auth token to this endpoint. [Read more](https://docs.saleor.io/docs/3.x/developer/extending/apps/installing-apps#installation-using-graphql-api)
 
-TODO - any other endpoints?
-
 ## Api handlers built-in SDK
 
 To hide Saleor internal logic, app-sdk provides handlers factories. They should work with minimal configuration, leaving
--- a/docs/api-handlers.md
+++ b/docs/api-handlers.md
@@ -1,0 +1,77 @@
+# Api Handlers
+
+Saleor Apps are meant to work in serverless environment, where Cloud Functions are the foundations of server-side code.
+
+Currently, Saleor heavily relies on Next.js, but in the future, other platforms will be supported.
+
+## Required handlers
+
+Saleor requires 2 endpoints to be available for a standalone app:
+
+- Manifest endpoint - Saleor asks for App information like name and permissions
+- Register endpoint - Saleor uses it to exchange authorization data
+
+TODO - any other endpoints?
+
+## Api handlers built-in SDK
+
+To hide Saleor internal logic, app-sdk provides handlers factories. They should work with minimal configuration, leaving
+App creators space for domain logic.
+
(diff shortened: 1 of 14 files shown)


docs/api-handlers.md
```diff
@@ -0,0 +1,77 @@
...
+Currently, Saleor heavily relies on Next.js, but in the future, other platforms will be supported.
+
+## Required handlers
+
+Saleor requires 2 endpoints to be available for a standalone app:
+
+- Manifest endpoint - Saleor asks for App information like name and permissions
+- Register endpoint - Saleor uses it to exchange authorization data
+
+TODO - any other endpoints?
```
lkostrowski, line 14: @krzysztofwolski can you confirm only manifest and register are required by Saleor?

src/handlers/next/create-manifest-handler.test.ts
```diff
@@ -0,0 +1,38 @@
...
+        };
+      },
+    });
+
+    const response = await handler(req, res);
+
+    /**
+     * TODO Fixme - why handler returns undefined?
+     */
+    expect(response).toEqual({});
```
lkostrowski, line 36: @zaiste can you help me why this test fails? Inside handler itself everything logs fine, but `Response.OK` seems to return undefined

tsup.config.ts
```diff
@@ -0,0 +1,20 @@
+import { defineConfig } from "tsup";
+
+export default defineConfig({
```
lkostrowski, line 3: Also extracted config from package.json and specify entry points. With upcoming updates we can changed imports to root package for easier moving files around later

docs/api-handlers.md
krzysztofwolski, line 14: Also the [appUrl](https://docs.saleor.io/docs/3.x/developer/extending/apps/manifest) which is used to display a view in the dashboard when app is opened. Not 100% sure if Saleor allow to create app without it 🤔

docs/api-handlers.md
```diff
@@ -0,0 +1,77 @@
...
+
+Saleor Apps are meant to work in serverless environment, where Cloud Functions are the foundations of server-side code.
+
+Currently, Saleor heavily relies on Next.js, but in the future, other platforms will be supported.
+
+## Required handlers
+
+Saleor requires 2 endpoints to be available for a standalone app:
+
+- Manifest endpoint - Saleor asks for App information like name and permissions
```
krzysztofwolski, line 11:
```suggestion
- Manifest endpoint - Returns JSON object with app properties, like its name or permissions. [Read more](https://docs.saleor.io/docs/3.x/developer/extending/apps/manifest)
```

docs/api-handlers.md
```diff
@@ -0,0 +1,77 @@
...
+- Register endpoint - Saleor uses it to exchange authorization data
```
krzysztofwolski, line 12:
```suggestion
- Register endpoint - During the installation process, Saleor sends `POST` request with auth token to this endpoint. [Read more](https://docs.saleor.io/docs/3.x/developer/extending/apps/installing-apps#installation-using-graphql-api)
```

krzysztofwolski (changes requested):

docs/api-handlers.md
lkostrowski: But appUrl is not endpoint, right? Its set by the factory itself

zaiste (approved):

krzysztofwolski (approved):

PR #64, opened 2022-10-04
Auth features

lkostrowski:
Move authorization HOC to SDK

### DIFF
--- a/src/app-bridge/use-dashboard-token.ts
+++ b/src/app-bridge/use-dashboard-token.ts
@@ -14,7 +14,7 @@ export interface DashboardTokenProps {
   tokenClaims: DashboardTokenPayload | null;
 }
 
-const debug = debugPkg.debug("app-sdk:AppBridge");
+const debug = debugPkg.debug("app-sdk:AppBridge:useDashboardToken");
 
 export const useDashboardToken = (): DashboardTokenProps => {
   const { appBridgeState } = useAppBridge();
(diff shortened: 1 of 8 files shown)


src/app-bridge/use-dashboard-token.ts
```diff
@@ -0,0 +1,49 @@
...
+  app: string;
+}
+
+export interface DashboardTokenProps {
+  isTokenValid: boolean;
+  hasAppToken: boolean;
+  tokenClaims: DashboardTokenPayload | null;
+}
+
+const debug = debugPkg.debug("app-sdk:AppBridge");
```
krzysztofwolski, line 17: Maybe this will be better when trying to find which function caused debug message (same as in middlewares)?

```suggestion
const debug = debugPkg.debug("app-sdk:AppBridge:useDashboardToken");
```

src/app-bridge/use-dashboard-token.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const useDashboardToken = (): DashboardTokenProps => {
+  const { appBridgeState } = useAppBridge();
+
+  const tokenClaims = useMemo(() => {
+    try {
+      if (appBridgeState?.token) {
+        debug("Trying to decode JWT token drom dashboard");
```
krzysztofwolski, line 25:
```suggestion
        debug("Trying to decode JWT token from dashboard");
```

src/app-bridge/with-authorization.tsx
```diff
@@ -0,0 +1,57 @@
...
+}
+
+type Props = {
+  unmounted?: ReactNode;
+  notIframe?: ReactNode;
+  noDashboardToken?: ReactNode;
+  dashboardTokenInvalid?: ReactNode;
+};
+
+export const withAuthorization =
```
krzysztofwolski, line 23: Not sure if thats a good message, but would like to have some anyway:

```suggestion
/**
* Render component only if it's requested from the dashboard environment.
**/
export const withAuthorization =
```

krzysztofwolski (changes requested):

PR #71, opened 2022-10-07
Add central SaleorApp instance

lkostrowski:
This introduces early SaleorApp class that should be an entry point for app setup and config

### DIFF
--- a/src/APL/vercel-apl.ts
+++ b/src/APL/vercel-apl.ts
@@ -1,7 +1,7 @@
 /* eslint-disable class-methods-use-this */
 import fetch, { Response } from "node-fetch";
 
-import { APL, AuthData } from "./apl";
+import { APL, AplReadyResult, AuthData } from "./apl";
 import { createAPLDebug } from "./apl-debug";
 
 const debug = createAPLDebug("VercelAPL");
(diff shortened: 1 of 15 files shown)


src/APL/vercel-apl.ts
```diff
@@ -122,4 +122,24 @@ export class VercelAPL implements APL {
...
+            `Env variable ${requiredEnvKey} doesnt exist or is empty. Ensure env variables exist`
+          ),
+        };
+      }
+    }
+
+    return {
+      ready: true,
+    };
+  }
```
krzysztofwolski, line 144: We can display all of them at once to avoid debugging variables one by one.

```suggestion
  async isReady(): Promise<AplReadyResult> {
    const missingVariables: string[] = [];
    for (const requiredEnvKey of Object.values(VercelAPLVariables)) {
      const envValue = process.env[requiredEnvKey];

      if (!envValue || envValue.length === 0) {
        missingVariables.push(requiredEnvKey)
      }
    }
    
    if (missingVariables.length > 0) {
        return {
          ready: false,
          error: new Error(
            `Env variables ${missingVariables} doesn't exist or is empty. Ensure env variables exist`
          ),
        };
      }

    return {
      ready: true,
    };
  }
```

krzysztofwolski (changes requested):

src/APL/vercel-apl.ts
lkostrowski: Good, I added custom error with stored errors (raw array) + message, so it can be easily parsed

PR #75, opened 2022-10-11
Disable token iss validation

lkostrowski:
Waiting for https://github.com/saleor/saleor/pull/10852 to complete. For now, iss check must be disabled, or apps are not working

### DIFF
--- a/src/APL/vercel-apl.ts
+++ b/src/APL/vercel-apl.ts
@@ -55,27 +55,28 @@ export type VercelAPLConfig = {
  *   - SALEOR_DEPLOYMENT_TOKEN: token for your particular Vercel deployment
  */
 export class VercelAPL implements APL {
-  private registerAppURL: string;
+  private registerAppURL?: string;
 
-  private deploymentToken: string;
+  private deploymentToken?: string;
 
   constructor(config?: VercelAPLConfig) {
     const registerAppURL =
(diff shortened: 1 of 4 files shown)


src/APL/vercel-apl.ts
```diff
@@ -55,27 +55,28 @@ export type VercelAPLConfig = {
...
       config?.registerAppURL || process.env[VercelAPLVariables.SALEOR_REGISTER_APP_URL];
-    if (!registerAppURL) {
-      throw new Error("Misconfiguration: please provide registerAppUrl");
-    }
+
     const deploymentToken =
       config?.deploymentToken || process.env[VercelAPLVariables.SALEOR_DEPLOYMENT_TOKEN];
-    if (!deploymentToken) {
-      throw new Error("Misconfiguration: please provide deploymentToken");
-    }
```
lkostrowski, line 72: @krzysztofwolski I have no better idea at the moment. Later the error will be thrown. Maybe we can do it like this, later we can think of something better. I don't want to overcomplicate API because of this

krzysztofwolski: Agree. The interface is not written in stone and since its not widely used yet we can roll with this approach and update it with v2 when we'll get more feedback

krzysztofwolski (approved):

PR #82, opened 2022-10-13
feat: add getLayout to return type of withAuthorization

krzysztofzuraw:
# What was done?
I've added `getLayout` to the return type of `withAuthorization`. Why? Because it turns out you need to apply `getLayout` after you wrap your component with `withAuthorization`. In the current situation TypeScript does not see `getLayout` on a wrapped component.

One example of such an error is [saleor-app-template](https://github.com/saleor/saleor-app-template/blob/main/pages/configuration.tsx#L133) - it turns out that the `Configuration` page will not have layout applied.

After a change in this PR following code will now work properly:

```tsx
function Component() {
  return <div>Component here</div>;
}

const authorizedComponent = withAuthorization()(Component);

authorizedComponent.getLayout = () => (
  <div>
    <Component />
  </div>
);
```

### DIFF
--- a/src/app-bridge/with-authorization.tsx
+++ b/src/app-bridge/with-authorization.tsx
@@ -27,6 +27,10 @@ const defaultProps: Props = {
   unmounted: <p>Loading</p>,
 };
 
+type WithAuthorizationHOC<P> = React.FunctionComponent<P> & {
+  getLayout?: (page: React.ReactElement) => React.ReactNode;
+};
+
 /**
  * Most likely, views from your app will be only accessibly inside Dashboard iframe.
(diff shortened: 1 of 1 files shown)


lkostrowski (approved):

krzysztofzuraw:
@lkostrowski what can I do to make `QA / test` check pass? 🤔

lkostrowski:
> @lkostrowski what can I do to make `QA / test` check pass? 🤔

I don't know what happened - looks like some error with github actions, but only related to code coverage. Maybe its somehow triggered from your "permissions" because it never happened before. I will merge it

PR #84, opened 2022-10-14
Encrypted metadata

### LINKED ISSUE #83: Encrypt private metadata
Apps should encrypt & decrypt data stored in private metadata (zero trust).

Possibly MetadataManager can use encryption logic. SDK can provide default one, but optional implementation should be possible to provide

krzysztofwolski:
Closes #83

Extend settings manager with encryption methods.

### DIFF
--- a/src/settings-manager/encrypted-metadata-manager.test.ts
+++ b/src/settings-manager/encrypted-metadata-manager.test.ts
@@ -1,0 +1,84 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+import {
+  DecryptCallback,
+  EncryptCallback,
+  EncryptedMetadataManager,
+} from "./encrypted-metadata-manager";
+import { MetadataEntry } from "./metadata-manager";
+
+const initialEntry = { key: "a", value: "without domain" };
+const encryptedEntry = {
+  key: "encrypted",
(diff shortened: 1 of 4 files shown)


src/settings-manager/metadata-manager.test.ts
```diff
@@ -1,28 +1,42 @@
...
 
 describe("settings-manager", () => {
   describe("metadata-manager", () => {
     const fetchMock = vi.fn(async () => metadata);
     const mutateMock = vi.fn(async (md: MetadataEntry[]) => [...metadata, ...md]);
+    const manager = new MetadataManager({
+      fetchMetadata: fetchMock,
+      mutateMetadata: mutateMock,
+      encryptionKey: "key",
+    });
```
lkostrowski, line 29: if metadata manager doesnt force encryption, we should not require encryption key, if someone will not use encryption

So I think its better to write it this way, WDYT

```
interface SettingsManager {
set(...);
get(...)
} 

class MetadataManager implements SettingsManager {} // or PlainMetadataManager

class EncryptedMetadataManager implements SettingsManager {} 

```

I think its quite clean

lkostrowski (changes requested):
Breaking SRP - lets split to 2 implementations of SettingsManager

lkostrowski:
@zaiste please check encryption impl

zaiste (approved):

lkostrowski (approved):

PR #85, opened 2022-10-14
Downgrade node-fetch to maintain compatibility with commonjs

mmiszy:
`node-fetch@3.x.x` dropped CommonJS support and requires ESM usage which is a huge pain in the neck. We can't use APL from `@saleor/app-sdk` in Checkout because of this.


mmiszy:
TBH node-fetch should be a peerDependency

lkostrowski (approved):

lkostrowski:
Released in 0.15.0
https://github.com/saleor/saleor-app-sdk/releases/tag/v0.15.0

we will make this peer dep 👍

PR #90, opened 2022-10-22
New Firestore APL in SDK

lkostrowski:
This PR adds new APL - for Google Firestore

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,4 @@
+# Config for testing FirestoreAPL, that requires actual connection to Firestore
+TEST_FIRESTORE_PROJECT_ID=
+TEST_FIRESTORE_PRIVATE_KEY=
+TEST_FIRESTORE_CLIENT_EMAIL=
--- a/docs/apl.md
+++ b/docs/apl.md
@@ -191,6 +191,7 @@ Or using environment variables: `UPSTASH_TOKEN`, `UPSTASH_URL`.
 It has a generous free tier and can be used both as an APL source and as a main app's database.
 
 SDK doesn't create a database - it requires a created Firestore client's collection to be provided in the constructor.
+SDK doesn't install this library - `@google-cloud/firestore` must be in `node_modules`.
 
 Check the [official Node.js client](https://www.npmjs.com/package/@google-cloud/firestore) to learn how to set up the database.
 
--- a/docs/apl.md
+++ b/docs/apl.md
@@ -200,7 +200,7 @@ import { FirestoreAPL, AuthData } from "@saleor/app-sdk/APL";
 import { CollectionReference, Firestore } from "@google-cloud/firestore"; // Must be installed!
 
 const fireStore = new Firestore(/* ...configuration */);
-const aplCollection = firestore.collection("apl"); // Collection name doesnt matter, APL uses reference
+const aplCollection = firestore.collection("apl"); // Collection name doesn't matter, APL uses reference
 
 const apl = new FirestoreAPL(aplCollection);
 ```
--- a/docs/apl.md
+++ b/docs/apl.md
@@ -196,10 +196,10 @@ Check the [official Node.js client](https://www.npmjs.com/package/@google-cloud/
 
 ```ts
 import { FirestoreAPL, AuthData } from "@saleor/app-sdk/APL";
-import { CollectionReference, Firestore } from "@google-cloud/firestore";
+import { CollectionReference, Firestore } from "@google-cloud/firestore"; // Must be installed!
 
 const fireStore = new Firestore(/* ...configuration */);
-const aplCollection = firestore.collection("apl") as CollectionReference<AuthData>; // Assert collection document type
+const aplCollection = firestore.collection("apl"); // Collection name doesnt matter, APL uses reference
 
 const apl = new FirestoreAPL(aplCollection);
 ```
--- a/docs/apl.md
+++ b/docs/apl.md
@@ -184,3 +184,22 @@ const apl = new UpstashAPL({
 ```
 
 Or using environment variables: `UPSTASH_TOKEN`, `UPSTASH_URL`.
+
+### Google Firestore APL
+
+[Google Firestore](https://cloud.google.com/firestore) is a modern noSQL database allowing real-time connections with short latency.
+It has a generous free tier and can be used both as an APL source and as a main app's database.
+
+SDK doesn't create a database - it requires a created Firestore client's collection to be provided in the constructor.
+
+Check the [official Node.js client](https://www.npmjs.com/package/@google-cloud/firestore) to learn how to set up the database.
+
+```ts
+import { FirestoreAPL, AuthData } from "@saleor/app-sdk/APL";
+import { CollectionReference, Firestore } from "@google-cloud/firestore";
+
+const fireStore = new Firestore(/* ...configuration */);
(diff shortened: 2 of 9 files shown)


.env.example
```diff
@@ -0,0 +1,4 @@
+# Config for testing FirestoreAPL, that requires actual connection to Firestore
+TEST_FIRESTORE_PROJECT_ID=
+TEST_FIRESTORE_PRIVATE_KEY=
+TEST_FIRESTORE_CLIENT_EMAIL=
```
lkostrowski, line 4: These are only for test purposes, so marked with TEST_

If they don't exist, tests will be skipped

package.json
```diff
@@ -33,19 +33,21 @@
     "uuid": "^8.3.2"
   },
   "devDependencies": {
-    "@vitest/coverage-c8": "^0.23.2",
-    "node-mocks-http": "^1.11.0",
-    "next": "^12.3.0",
+    "@google-cloud/firestore": "^6.4.0",
```
lkostrowski, line 36: this and dotenv are only added deps, rest was sorted

src/APL/firestore-apl.test.ts
```diff
@@ -0,0 +1,72 @@
...
+import { describe, expect, it } from "vitest";
+
+import { AuthData } from "./apl";
+import { FirestoreAPL } from "./firestore-apl";
+
+let firestore: Firestore | null = null;
+
+try {
+  firestore = new Firestore({
+    // @ts-ignore ignore import.meta, TODO
```
lkostrowski, line 11: I dont know how to set it. I dont want to mess with tsup, but also I can't configure it well. Changing modules in tsconfig either fails import.meta or fail deps imports.

@zaiste can you have a look, how should this be configured? Values *exist*, so probably we don't have to change tsconfig, but only add typings?

src/APL/firestore-apl.test.ts
```diff
@@ -0,0 +1,72 @@
...
+      private_key: import.meta.env.TEST_FIRESTORE_PRIVATE_KEY,
+      // @ts-ignore ignore import.meta, TODO
+      client_email: import.meta.env.TEST_FIRESTORE_CLIENT_EMAIL,
+    },
+  });
+} catch (e) {
+  // Omit test
+}
+
+describe.runIf(Boolean(firestore))("Firestore APL", async () => {
```
lkostrowski, line 24: I dont want to introduce complexty of writing e2e tests, and also Firestore is specific and hard to mock. Thats way these tests are opt in. If values are provided, they work. We should probably add them to CICD but don't require them locally

tsconfig.json
```diff
@@ -35,7 +35,9 @@
     // "paths": {},                                      /* Specify a set of entries that re-map imports to additional lookup locations. */
     // "rootDirs": [],                                   /* Allow multiple folders to be treated as one when resolving modules. */
     // "typeRoots": [],                                  /* Specify multiple folders that act like './node_modules/@types'. */
-    // "types": [],                                      /* Specify type package names to be included without being referenced in a source file. */
+    "types": [
+      "vite/client"
```
lkostrowski, line 39: it supports import.meta.env key

.env.example
krzysztofwolski: Maybe keeping them in the `.env.test` would be a better place then?

src/APL/firestore-apl.test.ts
krzysztofwolski: WDYT bout testing it with the Polly.js? 
- no need to create custom mocks since the real communication with the API is recorded and replayed
- updating the recording would be only required if a query to the Firestore API would change

lkostrowski: I will add separate issue for that, don't want to block it because of way we test (not sure how much time setup of polly will take)

krzysztofwolski (approved):
👍

lkostrowski:
Add example to prove that it works

lkostrowski:
@krzysztofwolski added few improvements + example of usage https://github.com/saleor/saleor-app-template/pull/116

src/APL/firestore-apl.test.ts
```diff
@@ -0,0 +1,77 @@
...
+const firestore = new Firestore({
+  // @ts-ignore ignore import.meta,
+  projectId: import.meta.env.TEST_FIRESTORE_PROJECT_ID,
+  credentials: {
+    // @ts-ignore ignore import.meta,
+    private_key: import.meta.env.TEST_FIRESTORE_PRIVATE_KEY,
+    // @ts-ignore ignore import.meta,
+    client_email: import.meta.env.TEST_FIRESTORE_CLIENT_EMAIL,
+  },
+});
```
lkostrowski, line 28: @zaiste can you help?

docs/apl.md
```diff
@@ -184,3 +184,23 @@ const apl = new UpstashAPL({
...
+SDK doesn't install this library - `@google-cloud/firestore` must be in `node_modules`.
+
+Check the [official Node.js client](https://www.npmjs.com/package/@google-cloud/firestore) to learn how to set up the database.
+
+```ts
+import { FirestoreAPL, AuthData } from "@saleor/app-sdk/APL";
+import { CollectionReference, Firestore } from "@google-cloud/firestore"; // Must be installed!
+
+const fireStore = new Firestore(/* ...configuration */);
+const aplCollection = firestore.collection("apl"); // Collection name doesnt matter, APL uses reference
```
krzysztofwolski, line 203:
```suggestion
const aplCollection = firestore.collection("apl"); // Collection name doesn't matter, APL uses reference
```

src/APL/firestore-apl.ts
```diff
@@ -0,0 +1,126 @@
...
+    try {
+      await this.isFirestoreConnected();
+
+      return {
+        configured: true,
+      };
+    } catch (e) {
+      return {
+        configured: false,
+        error: new Error("Firestore cant connect to server"),
```
krzysztofwolski, line 52:
```suggestion
        error: new Error("Firestore can't connect to server"),
```

src/APL/firestore-apl.ts
```diff
@@ -0,0 +1,126 @@
...
+  async getAll(): Promise<AuthData[]> {
+    debug("Attempt to get all auth data");
+
+    return this.firebaseCollection
+      .get()
+      .then((collection) => collection.docs.map((doc) => doc.data()));
+  }
+
+  /**
+   * This APL is always configured, because it accepts the Collection, which means Client is already defined.
```
krzysztofwolski, line 39:
```suggestion
   * This APL is always configured because it accepts the Collection, which means the Client is already defined.
```

src/APL/firestore-apl.ts
```diff
@@ -0,0 +1,126 @@
...
+    ]);
+
+    switch (timeoutResult) {
+      case "network": {
+        debug("Firestore responds from network");
+        return true;
+      }
+      case "timeout":
+      default: {
+        debug("Firestore cant connect to network in 500ms");
```
krzysztofwolski, line 121:
```suggestion
        debug("Firestore can't connect to network in 500ms");
```

krzysztofwolski (approved):
Besides the typos, LGTM

lkostrowski:
Closing for now, we can re-implement this if monorepo is there. This PR can still work as an example

Issue #92, opened 2022-10-24
🚀  app-sdk monorepo


lkostrowski:
SDK will likely need monorepo, because:

1. It provides code both for node and browser, having them in single package causes errors already (index file exposes fs, causing broken imports in browser)
2. It will likely expand with many connections to 3rd party and frameworks - Next.js specific, Remix specific, React specific

We can introduce monorepo with example packages:
1. sdk/core - typings and some basic symbols
2. sdk/next - Next adapters for handers
3. sdk/remix - like Next but for Remix
4. sdk/react - UI helpers
5. sdk/app-bridge - App bridge

etc

PR #98, opened 2022-10-25
Remove index export

lkostrowski:
SDK contains many packages, including node & browser only. Current index import doesn't even work for browser, because it imports node-only code and it breaks.

Hence lets remove it and use tree-shakable imports from specific entry point.

Maybe related for next iteration: #92 

Before merging, lets check if apps work, otherwise fix imports first

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,6 +120,11 @@
       "types": "./saleor-app.d.ts",
       "import": "./saleor-app.mjs",
       "require": "./saleor-app.js"
+    },
+    "./infer-webhooks": {
+      "types": "./infer-webhooks.d.ts",
+      "import": "./infer-webhooks.mjs",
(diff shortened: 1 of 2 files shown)


krzysztofwolski (changes requested):
Compared the output, seems all _intended_ packages are exported. Additional pro - test files seems to be no longer included in the bundle:
DTS dist/saleor-app.test.d.ts        1.00 B
DTS dist/setup-tests.d.ts            1.00 B
DTS dist/urls.test.d.ts              1.00 B)

When building template using this code had an issue with imports:
`import { SaleorApp } from "@saleor/app-sdk";`
Had to be changed to 
`import { SaleorApp } from "@saleor/app-sdk/saleor-app";`
Is this breaking change intentional?

PR #100, opened 2022-10-25
Add functions to process the Saleor webhooks

krzysztofwolski:


package.json
```diff
@@ -19,32 +19,34 @@
   "author": "",
   "license": "ISC",
   "peerDependencies": {
+    "next": "^12",
     "react": ">=17",
     "react-dom": ">=17",
-    "next": "^12"
+    "raw-body": ">=2.5.0"
```
lkostrowski, line 25: Why peer?

src/handlers/next/process-saleor-webhook.ts
```diff
@@ -0,0 +1,121 @@
...
+  MISSING_DOMAIN_HEADER,
+  MISSING_EVENT_HEADER,
+  MISSING_PAYLOAD_HEADER,
+  MISSING_SIGNATURE_HEADER,
+  MISSING_REQUEST_BODY,
+  WRONG_EVENT,
+  NOT_REGISTERED,
+  SIGNATURE_VERIFICATION_FAILED,
+  WRONG_METHOD,
+}
```
lkostrowski, line 23: I'd say that we should entirely avoid enums, but if there is any actual reason - they should be named (`MISSING_HOST_HEADER="MISSING_HOST_HEADER"`)

src/handlers/next/process-saleor-webhook.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export interface SaleorWebhookContext<T> {
```
lkostrowski, line 25: WebhookContext or EventContext?

src/handlers/next/process-saleor-webhook.ts
```diff
@@ -0,0 +1,121 @@
...
+  data?: { baseUrl: string; event: string; payload: T; authData: AuthData };
+  error?: SaleorWebhookError;
```
lkostrowski, line 27: it will console log to `error: 0`. Can be we use actual `class MissingDomainHeaderError extends Error`?

lkostrowski, line 27: This should be typed with union, because its always *either* data *or* error

```
{
  data?: { baseUrl: string; event: string; payload: T; authData: AuthData };
} |
{
  error?: SaleorWebhookError;
}
```

src/verify-signature.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  try {
+    await jose.flattenedVerify(jws, remoteJwks);
+    debug("JWKS verified");
+  } catch {
+    debug("JWKS verification failed");
+
+    // TODO: if we chose the way of making framework agnostic verify functions, we should unify
+    // how we return the errors
+    return "Verification using public key has failed.";
```
lkostrowski, line 30: We should start with returning Error or throwing Error. I can contain message but also can be handled agnostically

lkostrowski (changes requested):
Some changes, but I'd like to do another round on `createSaleorEventHandler`, I think it will provider better API.

Errors should not be a problem, because its signature can include:
```
handler() {/* This triggers when everything is ok */},
onError(error: WebhookError) { /* This triggers on error */}
```

package.json
krzysztofwolski: This package is required only in case of using nextjs helpers (createSaleorEventHandler use nextjs request types)

src/handlers/next/process-saleor-webhook.ts
krzysztofwolski: IMO since theres baseURL in the object, it describes the webhook.

src/handlers/next/process-saleor-webhook.ts
krzysztofwolski: Extended error class to contain extra data like error code and HTTP status code (unauthorized, wrong request, etc)

src/handlers/next/saleor-webhook.ts
```diff
@@ -0,0 +1,99 @@
...
+    this.asyncEvent = asyncEvent;
+    this.isActive = isActive;
+    this.apl = apl;
+  }
+
+  targetUrl(baseUrl: string) {
+    return new URL(this.webhookUrl, baseUrl).href;
+  }
+
+  getManifest(baseUrl: string): AppWebhook {
```
lkostrowski, line 63: `getWebhookManifest(baseUrl: string): WebhookManifest`

src/handlers/next/saleor-webhook.ts
```diff
@@ -0,0 +1,99 @@
...
+      isActive: this.isActive,
+      // Query can be provided as plaintext..
+      ...(this.query && { query: this.query }),
+      // ...GQL AST which has to be stringified..
+      ...(this.subscriptionQueryAst && { query: gqlAstToString(this.subscriptionQueryAst) }),
+      // or no query at all. In such case default webhook payload will be sent
+    };
+  }
+
+  handler(handlerFn: NextWebhookApiHandler<TPayload>): NextApiHandler {
```
lkostrowski, line 77: renamte to createHandler

src/verify-signature.ts
```diff
@@ -0,0 +1,33 @@
+import * as jose from "jose";
+
+import { createDebug } from "./debug";
+import { getJwksUrl } from "./urls";
+
+const debug = createDebug("verify-signature");
+
+export const verifySignature = async (domain: string, signature: string, rawBody: string) => {
```
lkostrowski, line 8: if this is extracted from middleware, remove that code and use this function there

lkostrowski (changes requested):

src/handlers/next/index.ts
```diff
@@ -1,2 +1,4 @@
 export * from "./create-app-register-handler";
 export * from "./create-manifest-handler";
+export * from "./process-async-saleor-webhook";
```
lkostrowski, line 3: how about removing this export

src/handlers/next/saleor-async-webhook.ts
```diff
@@ -0,0 +1,140 @@
...
+    this.isActive = isActive;
+    this.apl = apl;
+  }
+
+  /**
+   * Returns full URL to the webhook, based on provided baseUrl.
+   *
+   * @param baseUrl Base URL used by your application
+   */
+  targetUrl(baseUrl: string) {
```
lkostrowski, line 87: `getTargetUrl`

lkostrowski (changes requested):

src/handlers/next/process-async-saleor-webhook.test.ts
```diff
@@ -0,0 +1,147 @@
...
+    await expect(
+      processAsyncSaleorWebhook({
+        req: mockRequest,
+        apl: mockAPL,
+        allowedEvent: "PRODUCT_UPDATED",
+      })
+    ).rejects.toThrow("Missing saleor-signature header");
+  });
+
+  // TODO: Fix the test - not able to overwrite import mock
```
lkostrowski, line 110: Maybe try to restore mock first?
`vi.mocked(example.calc).mockRestore()`

https://vitest.dev/api/#vi-mocked

src/handlers/next/saleor-async-webhook.test.ts
```diff
@@ -0,0 +1,75 @@
+import { createMocks } from "node-mocks-http";
+import { describe, expect, it, vi } from "vitest";
+
+import { APL } from "../../APL";
+import { NextWebhookApiHandler, SaleorAsyncWebhook } from "./saleor-async-webhook";
+
+const webhookPath = "api/webhooks/product-updated";
```
lkostrowski, line 7: what happens if I add `/` at the beginning?

src/verify-signature.ts
```diff
@@ -0,0 +1,38 @@
...
+const debug = createDebug("verify-signature");
+
+/**
+ * Verify payload signature with public key of given `domain`
+ * https://docs.saleor.io/docs/3.x/developer/extending/apps/asynchronous-webhooks#payload-signature
+ *
+ * @param domain Saleor API domain
+ * @param signature Value of the saleor-signature header
+ * @param rawBody String value of request body
+ * @returns
```
lkostrowski, line 15: Lets remove params if they dont provide any value. In general typescript handles it. Only if param requires some deep explainer it makes sense to add it. IMO we can stick to general comment in JSDoc format

src/verify-signature.ts
krzysztofwolski: Used code in the existing middleware

src/handlers/next/saleor-async-webhook.test.ts
krzysztofwolski: Both will result in the same URL

src/handlers/next/get-base-url.ts
```diff
@@ -0,0 +1,9 @@
+import { NextApiRequest } from "next";
+
+export const getBaseURL = (req: NextApiRequest) => {
```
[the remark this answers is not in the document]
krzysztofwolski: Since the method works on the headers, moved it to the ./headers.ts and used it in the middleware.

src/index.ts
```diff
@@ -1,6 +1,9 @@
+import { AppWebhook as WebhookManifest } from "./types";
+
 export * from "./const";
 export * from "./headers";
 export * from "./infer-webhooks";
 export * from "./saleor-app";
 export * from "./types";
 export * from "./urls";
+export { WebhookManifest };
```
lkostrowski, line 9: index.ts is removed - rebase main branch

lkostrowski (changes requested):

lkostrowski (approved):

Issue #94, opened 2022-10-24
Create Webhook Handler factory


lkostrowski:
`createSaleorWebhookHandler` exported from `@saleor/app-sdk/handler/next`

```
// src/pages/api/webhooks/order_created.ts

export default createSaleorWebhookHandler({
  eventType: string | string[] 
  ...
})

```

```
// eg:
const handler: SaleorEventHandler<OrderCreatedDocument | OrderFullfilledDocument> = (req, res) => {
  const {apl, payload, domain, eventType} = req.context;

  // logic
}

export default createSaleorWebhookHandler({
  handler: handler
})
```

- [x] Check what Saleor event type
- [x] Inject SaleorApp / APL instance
- [x] Validate tokens and headers
- [x] Check payload checksum
- [x] Type extending Next's request with Saleors metadata, eg `SaleorEventRequest<PayloadGraphqlDocument>`
- [ ] Allow multiple event types
- [x] Documentation
- [x] Example

lkostrowski:
QA made on https://github.com/saleor/saleor-app-slack/pull/23

Here are problems:
1. Missing docs
2. When Webhook received empty query, installation failed, instead using some default payload
3. Wrong requests (eg empty body, wrong method) didn't receive error, error was somewhere silently handled

@krzysztofwolski

krzysztofwolski:
1. Docs PR: #109
2. Added in #107
3. Added separate ticket

PR #114, opened 2022-11-16
Add saleorApiUrl state param to AppBridge

lkostrowski:
This PR exposes access to new saleor-api-url header which is predictable URL for graphql

Ref #94

### DIFF
--- a/docs/app-bridge.md
+++ b/docs/app-bridge.md
@@ -35,13 +35,13 @@ type AppBridgeState = {
   token?: string;
   id: string;
   ready: boolean;
-  /**
-   * @deprecated - use saleorApiUrl instead
-   */
   domain: string;
   path: string;
   theme: ThemeType;
   locale: LocaleCode; // See src/locales.ts
+  /**
+   * Full URL including protocol and path where graphql is available
+   */
   saleorApiUrl: string;
 };
 ```
--- a/docs/app-bridge.md
+++ b/docs/app-bridge.md
@@ -17,6 +17,8 @@ Options object is following:
 ```
 type AppBridgeOptions = {
   targetDomain?: string;
+  saleorApiUrl?: string;
+  initialLocale?: LocaleCode;
 };
 ```
 
@@ -33,9 +35,14 @@ type AppBridgeState = {
   token?: string;
   id: string;
   ready: boolean;
+  /**
+   * @deprecated - use saleorApiUrl instead
+   */
   domain: string;
   path: string;
   theme: ThemeType;
+  locale: LocaleCode; // See src/locales.ts
+  saleorApiUrl: string;
 };
 ```
 
--- a/src/app-bridge/app-bridge.ts
+++ b/src/app-bridge/app-bridge.ts
@@ -3,6 +3,7 @@ import debugPkg from "debug";
 import { LocaleCode } from "../locales";
 import { Actions } from "./actions";
 import { AppBridgeState, AppBridgeStateContainer } from "./app-bridge-state";
+import { AppIframeParams } from "./app-iframe-params";
 import { SSR } from "./constants";
 import { Events, EventType, PayloadOfEvent, ThemeType } from "./events";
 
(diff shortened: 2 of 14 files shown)


docs/app-bridge.md
```diff
@@ -33,9 +35,14 @@ type AppBridgeState = {
...
   id: string;
   ready: boolean;
+  /**
+   * @deprecated - use saleorApiUrl instead
+   */
   domain: string;
   path: string;
   theme: ThemeType;
+  locale: LocaleCode; // See src/locales.ts
+  saleorApiUrl: string;
```
lkostrowski, line 45: @mmiszy is this exposed in AppBridgeState enough for you? Or what exactly are you trying to achieve?

At the moment `domain` is deprecated, so we have time to migrate older apps

przlada (approved):

src/urls.ts
```diff
@@ -4,9 +4,18 @@
...
   saleorDomain === "localhost:8000" ? "http" : "https";
 
+/**
+ * @deprecated use getJwksUrlFromSaleorApiUrl
+ */
 export const getJwksUrl = (saleorDomain: string): string =>
   `${resolveUrlProtocol(saleorDomain)}://${saleorDomain}/.well-known/jwks.json`;
 
+export const getJwksUrlFromSaleorApiUrl = (saleorApiUrl: string): string =>
+  `${new URL(saleorApiUrl).origin}/.well-known/jwks.json`;
```
lkostrowski, line 14: @przlada I assume path to jwks.json is always static

mmiszy: In the future we'll not need to do this shenanigans: https://github.com/saleor/saleor/issues/10800

src/app-bridge/app-bridge.ts
```diff
@@ -68,23 +69,30 @@ const createEmptySubscribeMap = (): SubscribeMap => ({
...
 
 /**
  * TODO: Probably remove empty string fallback
  */
-const getDomainFromUrl = () => new URL(window.location.href).searchParams.get("domain") || "";
+const getDomainFromUrl = () =>
+  new URL(window.location.href).searchParams.get(AppIframeParams.DOMAIN) || "";
+
+const getSaleorApiUrlFromUrl = () =>
+  new URL(window.location.href).searchParams.get(AppIframeParams.SALEOR_API_URL) || "";
```
mmiszy, line 90: Is there a need to provide both `?domain=…` and `?saleorApiUrl=…`? I think we should not require the `domain` param as it can be generated from `saleorApiUrl`.

mmiszy (changes requested):

src/app-bridge/app-bridge.ts
mmiszy: This will still error if only `saleorApiUrl` is provided:
https://github.com/saleor/saleor-app-sdk/blob/5b9e4bb2f5f068c8aa86f56d5375f78d17e72961/src/app-bridge/app-bridge.ts#L132-L136

lkostrowski: We can, but I wanted to make an addition, without breaking existing flows, because I'm not sure about the effort of existing apps.

My goal was to add extra param, on top of existing "domain", so new apps can use it now, old ones can slowly migrate 

Requiring "?domain" is contact between App and Dashboard only

I suggest that we make an addition, start using it and then create a migration guide

mmiszy: My suggestion would not break existing apps but allow new apps only to use `saleorApiUrl` which I believe is the end goal.

lkostrowski: I can do that. But technically for app it doesn't change, because its dashboard which provides both. App will just have one more field

lkostrowski: Done, now error is printed only if both params are missing

krzysztofwolski (approved):

PR #121, opened 2022-11-23
Add protected handlers for next

### LINKED ISSUE #94: Create Webhook Handler factory

### LINKED ISSUE #104: Create factory for handlers available only from the dashboard
Applications usually have endpoints to fetch/update app configuration used by the App interface inside of the Dashboard.

The factory should return handler validating incoming requests and their tokens.

Ref #94
Ref https://github.com/saleor/react-storefront/pull/592

krzysztofwolski:
Close #104 

Usage example can be found here: https://github.com/saleor/saleor-app-template/pull/119

### DIFF
--- a/src/handlers/next/create-protected-handler.ts
+++ b/src/handlers/next/create-protected-handler.ts
@@ -30,15 +30,12 @@ export type NextProtectedApiHandler<TResp = unknown> = (
 /**
  * Wraps provided function, to ensure incoming request comes from Saleor Dashboard.
  * Also provides additional `context` object containing request properties.
- *
- * @param handlerFn NextApiHandler function which takes additional `context` argument
- * @returns NextApiHandler
  */
 export const createProtectedHandler =
   (handlerFn: NextProtectedApiHandler, apl: APL): NextApiHandler =>
-  async (req, res) => {
+  (req, res) => {
     debug("Protected handler called");
-    await processSaleorProtectedHandler({
+    processSaleorProtectedHandler({
       req,
       apl,
     })
--- a/src/handlers/next/create-protected-handler.ts
+++ b/src/handlers/next/create-protected-handler.ts
@@ -1,0 +1,60 @@
+import { NextApiHandler, NextApiRequest, NextApiResponse } from "next";
+
+import { APL } from "../../APL";
+import { createDebug } from "../../debug";
+import { ProtectedHandlerContext } from "./process-async-saleor-webhook";
+import {
+  processSaleorProtectedHandler,
+  ProtectedHandlerError,
+  SaleorProtectedHandlerError,
+} from "./process-protected-handler";
+
+const debug = createDebug("ProtectedHandler");
+
+export const ProtectedHandlerErrorCodeMap: Record<SaleorProtectedHandlerError, number> = {
+  OTHER: 500,
+  MISSING_HOST_HEADER: 400,
+  MISSING_DOMAIN_HEADER: 400,
+  NOT_REGISTERED: 401,
+  JWT_VERIFICATION_FAILED: 401,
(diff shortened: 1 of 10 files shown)


src/handlers/next/create-protected-handler.ts
```diff
@@ -0,0 +1,60 @@
...
+  res: NextApiResponse<TResp>,
+  ctx: ProtectedHandlerContext
+) => unknown | Promise<unknown>;
+
+/**
+ * Wraps provided function, to ensure incoming request comes from Saleor Dashboard.
+ * Also provides additional `context` object containing request properties.
+ *
+ * @param handlerFn NextApiHandler function which takes additional `context` argument
+ * @returns NextApiHandler
```
lkostrowski, line 35: Lets remove autogenerated param jsdoc statements, they are redundant in typescript. Type NextApiHandler can be described with extra comment

src/handlers/next/create-protected-handler.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+export const createProtectedHandler =
+  (handlerFn: NextProtectedApiHandler, apl: APL): NextApiHandler =>
+  async (req, res) => {
+    debug("Protected handler called");
+    await processSaleorProtectedHandler({
```
lkostrowski, line 41: you are using async/await + then/catch. Maybe lets stick to one for simplicity?

src/handlers/next/process-protected-handler.ts
```diff
@@ -0,0 +1,106 @@
...
+    throw new ProtectedHandlerError("Missing saleor-domain header", "MISSING_DOMAIN_HEADER");
+  }
+
+  if (!token) {
+    debug("Missing authorization-bearer header");
+    throw new ProtectedHandlerError(
+      "Missing authorization-bearer header",
+      "MISSING_AUTHORIZATION_BEARER_HEADER"
+    );
+  }
```
lkostrowski, line 75: Not scope of this PR but maybe we should write some common assertions library, that will share this logic, eg

```
assert.authHeader(req); // void or throw
```

We will have unified errors which are technically the same, we can add context of which handler in extra argument

lkostrowski (approved):

src/handlers/next/create-protected-handler.ts
krzysztofwolski: Turns out it was not needed. Removed await

src/handlers/next/create-protected-handler.ts
krzysztofwolski: Removed 👍

src/handlers/next/process-protected-handler.ts
krzysztofwolski: Love the idea. Added issue #125

lkostrowski (approved):

PR #128, opened 2022-12-02
Change AppBridge interval to timeout

lkostrowski:
Looks like AppBridge was using `setInterval` instead `setTimeout`to timeout missed dispatched, this PR fixes it


src/app-bridge/app-bridge.ts
```diff
@@ -200,49 +200,50 @@ export class AppBridge {
...
-      } else {
-        debug("Calling window.parent.postMessage with %j", action);
-
-        window.parent.postMessage(
-          {
-            type: action.type,
-            payload: action.payload,
-          },
-          "*"
+        return;
```
lkostrowski, line 203: changing `else` nesting to early return

orzechdev (approved):

PR #129, opened 2022-12-02
Add action that informs Dashboard that AppBridge is ready

lkostrowski:
Context - when app is loaded it will be mounted in full height iframe, blinding with white/dark background (depending on app config) before theme is sent (after js loaded)

To avoid this situation, Dashboard should control when iframe is visible. Dashboard should show loading instead of iframe, and when app is loaded, show iframe

To achieve that, Dashboard must be informed when app is ready, hence new action `notifyReady` in SDK which is automatically sent when AppBridge is loaded (which means JS is loaded)

### DIFF
--- a/src/app-bridge/app-bridge.test.ts
+++ b/src/app-bridge/app-bridge.test.ts
@@ -3,7 +3,14 @@ import { beforeEach, describe, expect, it, vi } from "vitest";
 
 import { LocaleCode } from "../locales";
 // eslint-disable-next-line
-import { actions, AppBridge, DispatchResponseEvent, HandshakeEvent, ThemeEvent } from ".";
+import {
+  actions,
+  ActionType,
+  AppBridge,
+  DispatchResponseEvent,
+  HandshakeEvent,
+  ThemeEvent,
+} from ".";
 
 // mock document.referrer
 const origin = "http://example.com";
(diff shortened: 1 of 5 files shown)


src/app-bridge/app-bridge.test.ts
```diff
@@ -232,4 +252,14 @@ describe("AppBridge", () => {
...
   });
+
+  it("dispatches 'notifyReady' action when created", (done) => {
+    appBridge = new AppBridge();
+
+    window.addEventListener("message", (event) => {
+      if (event.data.type === ActionType.notifyReady) {
+        done();
+      }
+    });
```
orzechdev, line 263: Shouldn't event listener be defined first?
```suggestion
    window.addEventListener("message", (event) => {
      if (event.data.type === ActionType.notifyReady) {
        done();
      }
    });
    
    appBridge = new AppBridge();
```

orzechdev (approved):

src/app-bridge/app-bridge.test.ts
lkostrowski: Can be, but postMessage is on the event queue so it did wait :)

krzysztofwolski (approved):

PR #138, opened 2022-12-08
Add authenticated fetch

lkostrowski:
Will add docs when this is released (eg in RC), tested/added to template

### DIFF
--- a/src/app-bridge/fetch.ts
+++ b/src/app-bridge/fetch.ts
@@ -3,8 +3,12 @@ import { AppBridge } from "./app-bridge";
 import { useAppBridge } from "./app-bridge-provider";
 
 type HasAppBridgeState = Pick<AppBridge, "getState">;
+
+/**
+ * Created decorated window.fetch with headers required by app-sdk Next api handlers utilities
+ */
 export const createAuthenticatedFetch =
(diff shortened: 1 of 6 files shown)


src/app-bridge/fetch.ts
```diff
@@ -0,0 +1,31 @@
+import { SALEOR_AUTHORIZATION_BEARER_HEADER, SALEOR_DOMAIN_HEADER } from "../const";
+import { AppBridge } from "./app-bridge";
+import { useAppBridge } from "./app-bridge-provider";
+
+type HasAppBridgeState = Pick<AppBridge, "getState">;
+export const createAuthenticatedFetch =
```
krzysztofwolski, line 6: How next ssr would handle the lack of window? 

Would be a good idea to add docstring explaining that is client-side only

krzysztofwolski (changes requested):

src/app-bridge/fetch.ts
lkostrowski: Good catch. You can't use it because there is no AppBridge. I replaced window with global.

We probably should assume in `_app.tsx` that apps are client-side only, server context doesn't make sense. I can document this

krzysztofwolski (approved):

PR #148, opened 2023-01-13
Add optional protection to createRegisterHandler

lkostrowski:
Docs will be updated later


krzysztofwolski (changes requested):
Would consider:
- make allowSaleorUrls just a function instead of array of string | function. Less branches in our code and tests, easier to work with
- maybe instead of `allowSaleorUrls` better choice would be `saleorApiUrlValidationFn` a better choice? (or at least allowedSaleorUrls)

krzysztofwolski (approved):

PR #149, opened 2023-01-13
Fix CloudAPL response parsing

krzysztofwolski:


przlada (approved):

src/APL/saleor-cloud-apl.ts
```diff
@@ -25,6 +25,15 @@ const mapAuthDataToAPIBody = (authData: AuthData) => ({
...
 });
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const mapAPIResponseToAuthData = (response: any): AuthData => ({
+  appId: response.saleor_app_id,
+  domain: response.domain,
+  jwks: response.jwks,
+  saleorApiUrl: response.saleor_api_url,
+  token: response.token,
+});
```
lkostrowski, line 35: this should be wrapped with try/catch, because `any` is incorrect,  its actually `unknown`. Anyway try/catch can print proper error if something went wrong

PR #159, opened 2023-01-23
Handle missing encryption key for the EncryptedMetadataManager

krzysztofwolski:
If the key is missing:
- for dev: use placeholder value and show the warning
- for production: log the error and exit the application due to misconfiguration

### DIFF
--- a/docs/settings-manager.md
+++ b/docs/settings-manager.md
@@ -80,4 +80,5 @@ new EncryptedMetadataManager({
 });
 ```
 
-Warning: If encryption key won't be passed, the application will exit.
+> **Warning**
+> If encryption key won't be passed, the application will exit.
--- a/docs/settings-manager.md
+++ b/docs/settings-manager.md
@@ -66,3 +66,18 @@ const settings = new MetadataManager({
   mutateMetadata: (md) => mutateMetadata(client, md),
 });
 ```
+
+# EncryptedMetadataManager
+
+This manager encrypts add the layer of encryption for all the stored data.
+To operate correctly, the encryption key needs to be passed to the constructor:
+
+```ts
+new EncryptedMetadataManager({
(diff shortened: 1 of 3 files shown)


docs/settings-manager.md
```diff
@@ -66,3 +66,18 @@ const settings = new MetadataManager({
...
+
+```ts
+new EncryptedMetadataManager({
+  encryptionKey: process.env.SECRET_KEY, // secrets should be saved in the environment variables, never in the source code
+  fetchMetadata: () => fetchAllMetadata(client),
+  mutateMetadata: (metadata) => mutateMetadata(client, metadata),
+});
+```
+
+Warning: If encryption key won't be passed, the application will exit.
```
lkostrowski, line 83: you can use fancy syntax :)

https://github.com/community/community/discussions/16925

src/settings-manager/encrypted-metadata-manager.ts
```diff
@@ -71,7 +71,20 @@ export class EncryptedMetadataManager implements SettingsManager {
       fetchMetadata,
       mutateMetadata,
     });
-    this.encryptionKey = encryptionKey;
+    if (encryptionKey) {
+      this.encryptionKey = encryptionKey;
+    } else {
+      console.warn("Encrypted Metadata Manager secret key has not been set.");
+      if (process.env.NODE_ENV === "production") {
+        console.error("Can't start the application without the secret key.");
```
lkostrowski, line 79: I think the error should be thrown, so it can be caught by tooling etc, process.exit(1) is "an extra"

lkostrowski (approved):

src/settings-manager/encrypted-metadata-manager.ts
krzysztofwolski: Catching an error with the tooling is an important argument. Changed the implementation.

lkostrowski (approved):

PR #161, opened 2023-01-24
Permission check for the protected handler

krzysztofwolski:
Add an optional array of the required permissions to be check in protected handlers

### DIFF
--- a/src/has-permissions-in-jwt-token.ts
+++ b/src/has-permissions-in-jwt-token.ts
@@ -5,7 +5,7 @@ import { DashboardTokenPayload } from "./verify-jwt";
 const debug = createDebug("checkJwtPermissions");
 
 export const hasPermissionsInJwtToken = (
-  tokenData?: DashboardTokenPayload,
+  tokenData?: Pick<DashboardTokenPayload, "user_permissions">,
   permissionsToCheckAgainst?: AppPermission[]
 ) => {
   debug(`Permissions required ${permissionsToCheckAgainst}`);
--- a/src/has-permissions-in-jwt-token.ts
+++ b/src/has-permissions-in-jwt-token.ts
@@ -1,0 +1,36 @@
+import { createDebug } from "./debug";
+import { AppPermission } from "./types";
+import { DashboardTokenPayload } from "./verify-jwt";
+
+const debug = createDebug("checkJwtPermissions");
+
+export const hasPermissionsInJwtToken = (
+  tokenData?: DashboardTokenPayload,
+  permissionsToCheckAgainst?: AppPermission[]
+) => {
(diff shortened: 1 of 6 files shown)


src/has-permissions-in-jwt-token.ts
```diff
@@ -0,0 +1,36 @@
+import { createDebug } from "./debug";
+import { AppPermission } from "./types";
+import { DashboardTokenPayload } from "./verify-jwt";
+
+const debug = createDebug("checkJwtPermissions");
+
+export const hasPermissionsInJwtToken = (
```
lkostrowski, line 7: Missing test

lkostrowski (changes requested):

src/has-permissions-in-jwt-token.ts
```diff
@@ -0,0 +1,36 @@
...
+  tokenData?: DashboardTokenPayload,
```
lkostrowski, line 8:
```suggestion
  tokenData?: Pick<DashboardTokenPayload, 'user_permissions'>,
```

Interface Segregation -> in tests you don't need to provide unused field and this function should not require them if  its not using them

src/has-permissions.in-jwt-token.test.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import { hasPermissionsInJwtToken } from "./has-permissions-in-jwt-token";
+import { DashboardTokenPayload } from "./verify-jwt";
+
+describe("hasPermissionsInJwtToken", () => {
+  it("Pass if no required permissions, and user has none", async () => {
+    const tokenData: Pick<DashboardTokenPayload, "user_permissions"> = {
+      user_permissions: [],
+    };
+    await expect(hasPermissionsInJwtToken(tokenData)).toBeTruthy();
```
lkostrowski, line 11: avoid `toBeTruthy();` (maybe we can add eslint rule) -> toBeTruthy is doint type coercion so eg `!![]` will pass. This an lead to some unexpected false positives.

There is `toBe(true)` which is safe

src/verify-jwt.ts
```diff
@@ -1,21 +1,30 @@
...
 import { getJwksUrlFromSaleorApiUrl } from "./urls";
 
 const debug = createDebug("verify-jwt");
 
 export interface DashboardTokenPayload extends jose.JWTPayload {
   app: string;
+  user_permissions: AppPermission[];
 }
 
 export interface verifyJWTArguments {
```
lkostrowski, line 15: interface should be UpperCased

lkostrowski (approved):
comments non blocking 👍

lkostrowski:
@krzysztofwolski please release to npm after merge

PR #162, opened 2023-01-24
Allow setting initial theme from constructor in AppBridge

lkostrowski:


src/app-bridge/app-bridge.test.ts
```diff
@@ -253,13 +253,26 @@ describe("AppBridge", () => {
...
+
+    window.location.href = `${origin}?domain=${domain}&id=appid&theme=${themeToOverwrite}`;
+
+    expect(new AppBridge().getState().theme).toBe(themeToOverwrite);
+
+    window.location.href = currentLocationHref;
   });
+
+  it("dispatches 'notifyReady' action when enabled in constructor", () =>
+    new Promise<void>((res) => {
```
lkostrowski, line 269: according to vitest, this is correct syntax and they don't support jest's `done()` callback, so this test was never working before

src/app-bridge/app-bridge.ts
```diff
@@ -94,11 +95,24 @@ const getDomainFromUrl = () =>
...
+      return undefined;
+  }
+};
+
 const getDefaultOptions = (): AppBridgeOptions => ({
   targetDomain: getDomainFromUrl(),
   saleorApiUrl: getSaleorApiUrlFromUrl(),
   initialLocale: getLocaleFromUrl() ?? "en",
-  autoNotifyReady: true,
+  autoNotifyReady: false,
```
lkostrowski, line 114: changing to false, because Dashboard hasn't implemented it yet, so its throwing unknown action

PR #167, opened 2023-02-03
Add user email and permissions to app bridge

lkostrowski:


src/app-bridge/app-bridge-state.ts
```diff
@@ -10,6 +11,17 @@ export type AppBridgeState = {
...
+    /**
+     * Original permissions of the user that is using the app.
+     * *Not* the same permissions as the app itself.
+     *
+     * Can be used by app to check if user is authorized to perform
+     * domain specific actions
+     */
+    ownPermissions: Permission[];
+    email: string;
+  };
```
lkostrowski, line 24: @saleor/storefront-checkout FYI we will release such thing in next release

Thanks to that, you can easily verify if user using the app has permissions

Common use case:
A stuff member (non admin) can open the app to perform search index action in Algolia.
However, only admin (MANAGE_APPS permission) should be able to update Algolia tokens

Using `createProtectedHandler` you can provide require permission to the endpoint.
To show this on frontend (eg disabled form) you can check it easily with this feature now

Keep in mind that in such case there should be backend logic that validates token.
App (using apptoken) can update its own metadata. User token without "manage apps" can't.

So, to safely perform this operation you need to
1. Check `MANAGE_APPS` on the frontend (AppBridgeState), to give user feedback
2. Send form to API function, to check token again (token is validated and permission can't be passed)
3. API function will use AppToken to update its configuration, not user

cc @krzysztofwolski please verify what I said :)

src/app-bridge/app-bridge-state.ts
```diff
@@ -10,6 +11,17 @@ export type AppBridgeState = {
...
   saleorApiUrl: string;
+  user?: {
+    /**
+     * Original permissions of the user that is using the app.
+     * *Not* the same permissions as the app itself.
+     *
+     * Can be used by app to check if user is authorized to perform
+     * domain specific actions
+     */
+    ownPermissions: Permission[];
```
krzysztofwolski, line 22:
```suggestion
    permissions: Permission[];
```

If the permissions are inside the user object, prefix `own` IMO can be skipped.

Can we get access to the permissions of the app in the root level of the appBridge?

src/app-bridge/app-bridge-state.ts
krzysztofwolski: Well said :) 
Using the token from the app bridge allows us to perform actions on behalf of a particular Staff user:
- the Core will also check the permissions
- event logs (for example, timeline in the Order Events) will be able to display who performed the action and which application was used

src/app-bridge/app-bridge-state.ts
krzysztofwolski: Alternative approach - what if we expose the contents of the token? We'll be able to avoid mapping one object to another, which ould lead to confusion 

```
{
  "iat": 1669191845,
  "owner": "saleor",
  "iss": "demo.eu.saleor.cloud",
  "exp": 1669278245,
  "token": "sblFekZuBIGW",
  "email": "admin@example.com",
  "type": "thirdparty",
  "user_id": "VXNlcjoyMg==",
  "is_staff": true,
  "app": "QXBwOjI3NQ==",
  "permissions": [],
  "user_permissions": [
    "MANAGE_PAGE_TYPES_AND_ATTRIBUTES",
    "MANAGE_PRODUCT_TYPES_AND_ATTRIBUTES",
    "MANAGE_DISCOUNTS",
    "MANAGE_TRANSLATIONS",
    "MANAGE_CHECKOUTS",
    "HANDLE_CHECKOUTS",
    "MANAGE_ORDERS"
  ]
}
```

krzysztofwolski (changes requested):

src/app-bridge/app-bridge-state.ts
lkostrowski: I've changed to the suggested `permissions`.

We can add token too in the next PRs

krzysztofwolski (approved):

PR #170, opened 2023-02-07
Remove Vercel APL

lkostrowski:


krzysztofwolski (approved):

HazemKhaled:
We may need to rename it to environment variables APL, why not to keep simple stateless option for single tenant apps

lkostrowski:
@HazemKhaled this is quite complex, becuase by design Saleor is performing a handshake **during** app installation/register.

To enable single-tenant env APL, we need to extract this token. It is possible of course - you can `console.log` it during installation and hardcode it (you can write your own EnvAPL if you like).

However, we are discussing if we should add EnvAPL in SDK. This will make it "official" or "recommended" approach, but its not the best way to do it:
1. It requires token to leak
2. It will break if app rotates token
3. It will break when we implement updates flow for the app (which will rotate token).

There are also better approaches to achieve that, depending on the infrastructure.
You can use Vercel EdgeConfig, that will do the job without exposing the token

PR #173, opened 2023-02-14
Expose verify-jwks and remove legacy domain usage instead of api url

lkostrowski:
1. Removes legacy "domain" header usage for constructing calls to Saleor
2. Exposes JWKS verification functions in public exports


mmiszy (approved):

src/verify-signature.ts
```diff
@@ -63,6 +40,12 @@ export const verifySignatureFromApiUrl = async (
   }
 };
 
+/**
+ * Verify Webhook payload signature from provided jwks string.
```
krzysztofwolski, line 44:
```suggestion
 * Verify the Webhook payload signature from provided JWKS string.
```

src/middleware/with-webhook-signature-verified.ts
```diff
@@ -2,15 +2,18 @@ import crypto from "crypto";
...
-import { verifySignature } from "../verify-signature";
+import { verifySignatureFromApiUrl } from "../verify-signature";
 import { createMiddlewareDebug } from "./middleware-debug";
 
 const debug = createMiddlewareDebug("withWebhookSignatureVerified");
 
 const ERROR_MESSAGE = "Webhook signature verification failed:";
 
+/**
+ * TODO: Add test
```
krzysztofwolski, line 15: Maybe creating an issue would be a better choice than adding such comments?

krzysztofwolski (approved):

src/middleware/with-webhook-signature-verified.ts
lkostrowski: I don't think it adds noise. TODO comments are parsed pretty well by IDE and they can be tracked when we go back to the code, touch these files etc.

Of course, having issue for everything will be also good, but I'm afraid of having hundreds of issues in the backlog that will never be delivered. Issue for such small things will be "never" in the sprint, so we will never write this test.

But having such a small comment *near* the code will remind us every time we go there, and maybe when I add another fix/feature there, I will write missing test.

I know its far from perfect...

However I will add extra issue for this specific test, because its important to have it tested

krzysztofwolski: Makes sense :+1:

PR #175, opened 2023-02-14
Add more debug messages to SaleorCloudAPL

### LINKED ISSUE #166: Add more explicit logs (debug) to CloudAPL (what are responses from network)

lkostrowski:
Closes #166

### DIFF
--- a/src/APL/saleor-cloud-apl.ts
+++ b/src/APL/saleor-cloud-apl.ts
@@ -12,11 +12,13 @@ export type SaleorCloudAPLConfig = {
 const validateResponseStatus = (response: Response) => {
   if (response.status === 404) {
     debug("Auth data not found");
+    debug("%O", response);
 
     throw new Error("Auth data not found");
   }
   if (!response.ok) {
     debug("Response failed with status %s", response.status);
+    debug("%O", response);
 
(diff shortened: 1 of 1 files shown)


src/APL/saleor-cloud-apl.ts
```diff
@@ -132,10 +139,13 @@ export class SaleorCloudAPL implements APL {
       });
 
       debug(`Delete responded with ${response.status} code`);
-    } catch (error) {
-      debug("Error during deleting the data: %s", error);
+    } catch (error: any) {
```
krzysztofwolski, line 142:
```
  } catch (error) {
    debug("error");
    let message = "Unknown error"
    if (error instanceof Error) {
      message = error.message;
    }
```

WDYT?

krzysztofwolski (approved):

src/APL/saleor-cloud-apl.ts
lkostrowski: I can check `"message" in error`, but we can't rely on `instanceof` because in js you can throw anything.

I will make it without `any`

lkostrowski: Refactored

PR #183, opened 2023-02-23
Add hooks to createAppRegsiterHandler

lkostrowski:
- [x] Add tests
- [x] Add docs
- [x] Add changeset

### DIFF
--- a/src/handlers/next/create-app-register-handler.test.ts
+++ b/src/handlers/next/create-app-register-handler.test.ts
@@ -122,7 +122,7 @@ describe("create-app-register-handler", () => {
         apl: mockApl,
         onRequestStart: mockOnRequestStart,
         onRequestVerified: mockOnRequestVerified,
-        onAuthAplFailed: mockOnAuthAplFailed,
+        onAplSetFailed: mockOnAuthAplFailed,
         onAuthAplSaved: mockOnAuthAplSaved,
       });
 
@@ -186,7 +186,7 @@ describe("create-app-register-handler", () => {
 
       const handler = createAppRegisterHandler({
         apl: mockApl,
-        onAuthAplFailed: mockOnAuthAplFailed,
+        onAplSetFailed: mockOnAuthAplFailed,
         onAuthAplSaved: mockOnAuthAplSaved,
       });
 
@@ -213,13 +213,20 @@ describe("create-app-register-handler", () => {
     });
 
     it("Allows to send custom error response via callback hook", async () => {
-      const mockOnRequestStart = vi.fn().mockImplementation((req, context: any) => {
-        throw context.respondWithError({
-          status: 401,
-          body: "test",
-          message: "test message",
-        });
-      });
+      const mockOnRequestStart = vi.fn().mockImplementation(
+        (
+          req,
+          context: {
+            respondWithError(params: { status: number; body: string; message: string }): Error;
+          }
+        ) => {
+          throw context.respondWithError({
+            status: 401,
+            body: "test",
+            message: "test message",
+          });
+        }
+      );
 
       const { res, req } = createMocks({
         /**
--- a/src/handlers/next/create-app-register-handler.test.ts
+++ b/src/handlers/next/create-app-register-handler.test.ts
@@ -1,15 +1,18 @@
 import { createMocks } from "node-mocks-http";
-import { describe, expect, it, vi } from "vitest";
+import { describe, expect, it, Mock, vi } from "vitest";
 
-import { APL } from "../../APL";
+import { APL, AuthData } from "../../APL";
 import { createAppRegisterHandler } from "./create-app-register-handler";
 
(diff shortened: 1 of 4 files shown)


src/handlers/next/create-app-register-handler.ts
```diff
@@ -22,6 +22,44 @@ export type CreateAppRegisterHandlerOptions = HasAPL & {
...
+      context: {
+        authToken: string;
+        saleorDomain: string;
+        saleorApiUrl: string;
+        appId: string;
+      }
+    ): Promise<void>;
+    onAuthAplSaved?(
+      request: Request,
+      context: {
```
krzysztofwolski, line 45: Maybe it should be a `AuthData` since it should be created now?

src/handlers/next/create-app-register-handler.ts
```diff
@@ -22,6 +22,44 @@ export type CreateAppRegisterHandlerOptions = HasAPL & {
    * or a function that receives a full Saleor API URL ad returns true/false.
    */
   allowedSaleorUrls?: Array<string | ((saleorApiUrl: string) => boolean)>;
+  hooks?: {
```
krzysztofwolski, line 25: Not sure if those args should be nested. Probably would prefer to keep it flat.

src/handlers/next/create-app-register-handler.ts
```diff
@@ -32,13 +70,22 @@ export type CreateAppRegisterHandlerOptions = HasAPL & {
...
   allowedSaleorUrls,
+  hooks,
 }: CreateAppRegisterHandlerOptions) => {
   const baseHandler: Handler = async (request) => {
     debug("Request received");
     const authToken = request.params.auth_token;
     const saleorDomain = request.headers[SALEOR_DOMAIN_HEADER] as string;
     const saleorApiUrl = request.headers[SALEOR_API_URL_HEADER] as string;
 
+    if (hooks?.onRequestStart) {
```
krzysztofwolski, line 81: Try/catch on calling the hooks?

src/handlers/next/create-app-register-handler.ts
```diff
@@ -112,16 +168,38 @@ export const createAppRegisterHandler = ({
         appId,
         jwks,
       });
-    } catch {
+
+      if (hooks?.onAuthAplSaved) {
```
krzysztofwolski, line 172: What if this callback fails?

src/handlers/next/create-app-register-handler.ts
```diff
@@ -13,6 +14,39 @@ import { validateAllowSaleorUrls } from "./validate-allow-saleor-urls";
 
 const debug = createDebug("createAppRegisterHandler");
 
+type HookCallbackErrorParams = {
+  status?: number;
+  body?: object;
+  message?: string;
+};
+
+class HookCallbackError extends Error {
```
krzysztofwolski, line 23:
```suggestion
class RegisterCallbackError extends Error {
```
Dont have a strong preference, but probably this name would be more clear for me if seen the first time in the logs.

krzysztofwolski: Especially if fn to create one is called createCallbackError

src/handlers/next/create-app-register-handler.ts
```diff
@@ -22,6 +56,49 @@ export type CreateAppRegisterHandlerOptions = HasAPL & {
...
+      respondWithError: typeof createCallbackError;
+    }
+  ): Promise<void>;
+  /**
+   * Run after APL fails to set AuthData
+   */
+  onAuthAplFailed?(
+    request: Request,
+    context: {
+      authData: AuthData;
```
krzysztofwolski, line 97: If the APL has failed, are the authData available?

src/handlers/next/create-app-register-handler.test.ts
```diff
@@ -90,4 +93,160 @@ describe("create-app-register-handler", () => {
...
+          authData: expectedAuthData,
+          error: expect.objectContaining({
+            message: "test error",
+          }),
+        })
+      );
+    });
+
+    it("Allows to send custom error response via callback hook", async () => {
+      const mockOnRequestStart = vi.fn().mockImplementation((req, context: any) => {
```
krzysztofwolski, line 216:
```suggestion
      const mockOnRequestStart = vi.fn().mockImplementation((req, context: unknown) => {
```
Or add ignore the comment. Since it's test code IMO it's allowed.

krzysztofwolski (changes requested):

src/handlers/next/create-app-register-handler.ts
lkostrowski: >If the APL has failed, are the authData available?

Yes, this is argument for `set` so you can still receive it

krzysztofwolski (approved):

PR #201, opened 2023-02-28
Improve register handler errors API

lkostrowski:


src/handlers/next/create-app-register-handler.ts
```diff
@@ -16,33 +16,49 @@ const debug = createDebug("createAppRegisterHandler");
...
+};
+
+export type RegisterHandlerResponseBody = {
+  success: boolean;
+  error?: {
+    code?: string;
+    message?: string;
+  };
+};
+export const createRegisterHandlerResponseBody = (
```
krzysztofwolski, line 45: Debug message in this fn could be helpfull. WDYT?

krzysztofwolski (approved):

src/handlers/next/create-app-register-handler.ts
lkostrowski: Really? It only does simple mapping and this error goes directly to the response, I don't see a benefit really

PR #213, opened 2023-03-12
Add EnvAPL

lkostrowski:
```[tasklist]
### TODO
- [ ] Docs
- [x] Test in app-template
```

Proof of concept:
https://github.com/lkostrowski/test-env-apl
https://github.com/lkostrowski/test-env-apl/blob/main/src/saleor-app.ts

First installation under the flag prints secrets

Zrzut ekranu 2023-03-14 o 10 38 36

Envs set in vercel:

Zrzut ekranu 2023-03-14 o 10 48 57

POC fetches APL from cloud function and prints it on frontend

Zrzut ekranu 2023-03-14 o 10 49 10

### DIFF
--- a/src/APL/apl.ts
+++ b/src/APL/apl.ts
@@ -1,9 +1,9 @@
 export interface AuthData {
-  domain: string;
+  domain?: string;
   token: string;
   saleorApiUrl: string;
   appId: string;
-  jwks: string;
+  jwks?: string;
 }
 
 export type AplReadyResult =
--- a/src/APL/env-apl.ts
+++ b/src/APL/env-apl.ts
@@ -1,0 +1,92 @@
+import { APL, AuthData } from "./apl";
+import { createAPLDebug } from "./apl-debug";
+
+const debug = createAPLDebug("EnvAPL");
+
+type AuthDataRequired = Omit<AuthData, "jwks" | "domain">;
+
+type Options = {
+  env: Record<keyof AuthDataRequired, string>;
+  /**
+   * Enable to log auth data to stdout.
+   * Do it once to save data in ENV and disable it later.
+   */
+  printAuthDataOnRegister?: boolean;
+};
+
+export class EnvApl implements APL {
+  private defaultOptions: Partial<Options> = {
+    printAuthDataOnRegister: false,
+  };
+
+  options: Options;
+
+  constructor(options: Options) {
+    if (!this.isAuthDataValid(options.env)) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        "EnvAPL constructor now filled with valid AuthData config. Try to install the app with \"printAuthDataOnRegister\" enabled and check console logs"
+      );
+    }
+
+    this.options = {
+      ...this.defaultOptions,
+      ...options,
+    };
+  }
+
+  private isAuthDataValid(authData: AuthData): boolean {
+    const keysToValidateAgainst: Array<keyof AuthData> = ["appId", "saleorApiUrl", "token"];
+
+    return keysToValidateAgainst.every(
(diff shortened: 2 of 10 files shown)


lkostrowski:
cc @mmiszy 

while this is not recommended (apps using this are hard to maintain or update), we can still have it in app-sdk for very simple use cases of single-tenant apps

please take a look if this implementation can work for checkout

src/APL/apl.ts
```diff
@@ -1,9 +1,9 @@
 export interface AuthData {
-  domain: string;
+  domain?: string;
   token: string;
   saleorApiUrl: string;
   appId: string;
-  jwks: string;
+  jwks?: string;
```
krzysztofwolski, line 6: Good idea

src/APL/env-apl.ts
```diff
@@ -0,0 +1,109 @@
...
+      ...this.defaultOptions,
+      ...options,
+    };
+  }
+
+  private isAuthDataValid(authData: AuthData): boolean {
+    const keysToValidateAgainst: Array<keyof AuthData> = ["appId", "saleorApiUrl", "token"];
+
+    return keysToValidateAgainst.every(
+      (key) => authData[key] && typeof authData[key] === "string" && authData[key]!.length > 0
```
krzysztofwolski, line 42: Wonder if it will work with the vercel? 
https://nextjs.org/docs/basic-features/environment-variables#loading-environment-variables

krzysztofwolski (approved):
Please make sure env validation wors on the vercel before the release

src/APL/env-apl.ts
lkostrowski: Can you elaborate? Vercel exposes env variables in `process.env.XYZ` but this APL doesn't even access it, everything is mapped in constructore

krzysztofwolski: You are right, constructor will take of this :bow:

krzysztofwolski: My concern was related to the way env vars are accessed, you have to do this the right way so it will be accessible by next-magic:
```
Note: In order to keep server-only secrets safe, environment variables are evaluated at build time, so only environment variables actually used will be included. This means that process.env is not a standard JavaScript object, so you’re not able to use [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment). Environment variables must be referenced as e.g. process.env.PUBLISHABLE_KEY, not const { PUBLISHABLE_KEY } = process.env.
```

lkostrowski: Yes, thats why I decided not to read env inside APL, because it could get pruned by webpack. Mapping envs happen outside of SDK so it should be safe for Vercel/Next

Of course I will test it and attach it here

lkostrowski:
@krzysztofwolski added readme

src/APL/env-apl.ts
```diff
@@ -0,0 +1,108 @@
...
+      configured: true,
+    } as const;
+  }
+
+  async set(authData: AuthData) {
+    if (this.options.printAuthDataOnRegister) {
+      // eslint-disable-next-line no-console
+      console.log("Displaying registration values for the app. Use them to configure EnvAPL");
+      // eslint-disable-next-line no-console
+      console.table(authData);
```
krzysztofwolski, line 72: TBH not a fan of the table - Usually its formatting falls apart when line length is right and it makes it harder to parse (for example in scripts).
Not mandatory request, leave decision to you :)

krzysztofwolski (approved):
Code looks all right, one consideration in the comment

src/APL/env-apl.ts
lkostrowski: Good point, in Vercel screenshot its broken

PR #224, opened 2023-03-31
Require only partial of req - headers - in processProtectedHandler

lkostrowski:


src/handlers/next/process-protected-handler.ts
```diff
@@ -49,6 +49,8 @@ type ProcessAsyncSaleorProtectedHandler = (
 /**
  * Perform security checks on given request and return ProtectedHandlerContext object.
  * In case of validation issues, instance of the ProtectedHandlerError will be thrown.
+ *
+ * Can pass entire next request or Headers with saleorApiUrl and token
```
peelar, line 53: Can it? Because the type only picks the `headers` key:

```ts
Pick<NextApiRequest, "headers">;
```

lkostrowski: Yes, Pick is a subset, so everything that contains `{headers: Headers}` will pass.

peelar: I am not sure I understand. I created [a ts playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAShCOUC8UDeAoKWoAsIEMATCAJwGcAuKM4EgSwDsBzAbk2wDMB7Lqm+5mwC+6dKEhQAKhBrIoABToBjANYAeOPAA0UAER4ipMroB8opVwaz8VabJQZsUAEb4SVffgA2XXVvZYBsTkHn7oQkA) to explain what I mean.

lkostrowski: This is how it works

https://www.typescriptlang.org/play?#code/C4TwDgpgBAShCOUC8UDeAoKWoAsIEMATCAJwGcAuKM4EgSwDsBzAbk2wDMB7Lqm+5mwC+6dKEhQAKhBrIoABToBjANYAeOPAA0UAER4ipMroB8opVwaz8SpRDDBpslAApgM4FSfAAlGhHoFlbAUCQIVJpyGNi4BMTkVADk7jSJWuxY3LxQiVmJ6AE2dg7eLmHwPkA

peelar: Ah, I get it! Thanks for the explanation.

peelar (approved):

.changeset/tough-tips-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+"@saleor/app-sdk": patch
+---
+
+processProtectedHandler no longer requires full NextApiRequest as an argument. It only needs headers now to check. Thanks to that, some requests like HTML <form> with tokens in BODY can be validated. Till now only fetch/ajax calls could have been validated
```
krzysztofwolski, line 5:
```suggestion
processProtectedHandler no longer requires a full NextApiRequest object as an argument. Now only the `headers` property is required to satisfy the type safety.

Thanks to that, some requests like HTML <form> with tokens in BODY can be validated. Till now only fetch/ajax calls could have been validated
```

krzysztofwolski (approved):

PR #227, opened 2023-04-11
Support comma-delimited x-forwarded-proto

mmiszy:
Some tunnel providers send multiple comma-separated values in `x-forwarded-proto` header. For example, in the case of ngrok it's `https,http`.

This PR introduces changes to handle such values correctly.


lkostrowski (approved):
Thanks

PR #228, opened 2023-04-12
Fix UpstashAPL serialization of the AuthData

krzysztofwolski:
Fix missing serialization of the whole request send to the Upstash

### DIFF
--- a/.changeset/eight-wombats-drop.md
+++ b/.changeset/eight-wombats-drop.md
@@ -1,0 +1,5 @@
+---
+"@saleor/app-sdk": patch
+---
+
+Fix serialization of the nested values in the UpstashAPL.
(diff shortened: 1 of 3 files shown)


lkostrowski (changes requested):
1. Please write a unit test for `process-saleor-webhook.ts` that will test scenario with previous, JSON format, it should handle this and overwrite with base64 without need to reinstall the app
2. ensure that other places, like register handler, are not affected

krzysztofwolski:
@lkostrowski I've took a look at Upstash npm Library and found out what was wrong ealier - the serialization was applied at JWKS only, not the whole request. After the change, it works now as intended. The format of the stored data did not changed (plain JSON). So the previously stored data should be still ok.
Seems the issue was present since we added JWKS keys to the AuthData.

lkostrowski (approved):

PR #241, opened 2023-05-16
Add delete method do settings manager and implement it metadata manager

lkostrowski:


src/settings-manager/metadata-manager.test.ts
```diff
@@ -53,5 +55,69 @@ describe("settings-manager", () => {
...
+        await manager.delete([
+          { key: "test1", domain: "https://test.com/graphql/" },
+          { key: "test2", domain: "https://test.com/graphql/" },
+        ]);
+
+        expect(deleteMetadataMock).toBeCalledTimes(2);
+        expect(deleteMetadataMock).toHaveBeenNthCalledWith(1, ["test__https://test.com/graphql/"]);
+        expect(deleteMetadataMock).toHaveBeenNthCalledWith(2, [
+          "test1__https://test.com/graphql/",
+          "test2__https://test.com/graphql/",
```
krzysztofwolski, line 118:
```suggestion
          { key: "test1", domain: "https://example.com/graphql/" },
          { key: "test2", domain: "https://example.com/graphql/" },
        ]);

        expect(deleteMetadataMock).toBeCalledTimes(2);
        expect(deleteMetadataMock).toHaveBeenNthCalledWith(1, ["test__https://example.com/graphql/"]);
        expect(deleteMetadataMock).toHaveBeenNthCalledWith(2, [
          "test1__https://example.com/graphql/",
          "test2__https://example.com/graphql/",
```
Not a big deal here, but would rather use `example.com` or our own domain for tests in case fetch client is not mocked - we dont want to ping some random domain from CI.

krzysztofwolski (changes requested):

krzysztofwolski (approved):

PR #265, opened 2023-07-21
Expose headers file. Added AppPermissions extration into AppBridge

lkostrowski:
1. Can now directly get app permissions from appbridge (some apps had to do that manually)
2. Can now import getBaseUrl from SDK, which was copy-pasted into apps


src/util/extract-app-permissions-from-jwt.ts
```diff
@@ -0,0 +1,9 @@
+import * as jose from "jose";
+
+import { AppPermission } from "../../dist/types";
+
+export const extractAppPermissionsFromJwt = (jwtToken: string): AppPermission[] => {
+  const tokenDecoded = jose.decodeJwt(jwtToken);
+
+  return tokenDecoded.permissions as AppPermission[];
```
peelar, line 8: **question**: Is there no need to parse it? Can we safely cast the type?

peelar (approved):

src/util/extract-app-permissions-from-jwt.ts
lkostrowski: Saleor guarantees that, if Saleor can't do that right, app will break anyway

PR #280, opened 2023-09-04
Mark query in manifest required

lkostrowski:


lkostrowski:
@krzysztofwolski I checked E&M and it looks fine 

```
export const notifyWebhook = new SaleorAsyncWebhook<NotifySubscriptionPayload>({
  name: "notify",
  webhookPath: "api/webhooks/notify",
  asyncEvent: "NOTIFY_USER",
  apl: saleorApp.apl,
  query: "{}", // We are using the default payload instead of subscription
});
```

it defines query as string which is ok for the new SDK

.changeset/tender-news-do.md
```diff
@@ -0,0 +1,9 @@
+---
+"@saleor/app-sdk": patch
+---
+
+In AppManifest, webhook query field is not required. Saleor rejects empty query, so Manifest was wrong before with with field optional
+
+This is technically a breaking change, but app without a query specified couldn't be installed in Saleor.
+
+Now Typescript will early show the error
```
peelar, line 9: It doesn't say what the change is

peelar (changes requested):

.changeset/tender-news-do.md
lkostrowski: rewrtitten

peelar (approved):

PR #285, opened 2023-09-08
feat: :sparkles: add `examples` directory with .gitmodules

peelar:
https://github.com/saleor/app-examples/pull/11

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,18 @@
+[submodule "examples/dashboard-extensions"]
+	path = examples/dashboard-extensions
+	url = https://github.com/saleor/saleor-app-template
+[submodule "examples/metadata-manager"]
+	path = examples/metadata-manager
+	url = https://github.com/saleor/saleor-app-template
+[submodule "examples/external-webhook"]
+	path = examples/external-webhook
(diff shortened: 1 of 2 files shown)


lkostrowski:
@peelar in general I like the idea, but I don't like that coupling

SDK depends on app-template, that contains SDK usage, on app-template branches, to mention this in SDK repo.

I think we should somehow move app-template code from app-template to app-sdk and keep them entirely here. Maybe submodules, maybe not, just I'd not decouple them.

lkostrowski (changes requested):
Please figure out decoupling from app-template repo

lkostrowski:
@peelar please hand over to @saleor/dex team

mmiszy:
Do we even need app-sdk examples if apps are open source already?

lkostrowski:
> Do we even need app-sdk examples if apps are open source already?

IMO yes but if DX team decides not to have them, we can remove them. 

I don't think apps are good examples because apps are complex and contain a lot of logic that hides what comes from SDK. Examples are just barebones app-template + app-sdk method applied on top. Very visible.

If we don't want to have examples to maintain, we should probably add more docs with examples written as snippets. Or - a code sandbox?

PR #296, opened 2023-10-06
Throw error if cloud apl GET fails

lkostrowski:
Test with existing app first

### DIFF
--- a/src/APL/saleor-cloud/saleor-cloud-apl.ts
+++ b/src/APL/saleor-cloud/saleor-cloud-apl.ts
@@ -24,6 +24,11 @@ export type GetAllAplResponseShape = {
   results: CloudAPLAuthDataShape[];
 };
 
+export const CloudAplError = {
+  FAILED_TO_REACH_API: "FAILED_TO_REACH_API",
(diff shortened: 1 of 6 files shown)


lkostrowski:
@saleor/delivery-engineering-js bump - maybe @peelar or @krzysztofwolski you can try to install this from PR to one of ongoing apps to just check if it doesn't break anything, before we merge?

src/APL/saleor-cloud/saleor-cloud-apl.ts
```diff
@@ -102,25 +97,32 @@ export class SaleorCloudAPL implements APL {
...
+      debug("No auth data for given saleorApiUrl");
       return undefined;
     }
 
     const parsedResponse = (await response.json().catch((e) => {
       debug("Failed to parse response: %s", extractErrorMessage(e));
       debug("%O", e);
+
+      throw new SaleorCloudAplError(
+        "RESPONSE_BODY_INVALID",
```
witoszekdev, line 123: Maybe these error codes should be an enum that we export from the library, so that it's easier for an app to handle these errors?

PR #309, opened 2023-11-28
Add OTEL for some longer sdk jobs

lkostrowski:


src/APL/saleor-cloud/saleor-cloud-apl.ts
```diff
@@ -95,79 +103,172 @@ export class SaleorCloudAPL implements APL {
...
+          [SemanticAttributes.PEER_SERVICE]: OTEL_APL_SERVICE_NAME,
+        },
+        kind: SpanKind.CLIENT,
+      },
+      async (span) => {
+        const response = await fetch(this.getUrlForDomain(saleorApiUrl), {
+          method: "GET",
+          headers: { "Content-Type": "application/json", ...this.headers },
+        }).catch((error) => {
+          debug("Failed to reach API call:  %s", extractErrorMessage(error));
```
witoszekdev, line 120: Why not use

```js
try {
  await fetch()
} catch (error) {
  // ...
}
```

src/APL/saleor-cloud/saleor-cloud-apl.ts
```diff
@@ -95,79 +103,172 @@ export class SaleorCloudAPL implements APL {
...
+          debug("%O", e);
+
+          span.recordException(`Failed to reach API call:  ${extractErrorMessage(e)}`);
+          span
+            .setStatus({
+              code: SpanStatusCode.ERROR,
+            })
+            .end();
+
+          throw new Error(`Error during saving the data: ${extractErrorMessage(e)}`);
```
witoszekdev, line 257: Shouldn't `SaleorCloudAplError` be used here instead of the generic `Error` class?

src/APL/saleor-cloud/saleor-cloud-apl.ts
lkostrowski: 1. Its the same
2. It was written like this before, in this PR I only wrapped entire function with `startActiveSpan`, so I don't see value of refactoring at the same time (and I also don't see problem with this syntax)

If we agree that try/catch adds some value, we should probably set this on ESLint level as a rule

src/APL/saleor-cloud/saleor-cloud-apl.ts
lkostrowski: Good catch. Out of scope, but I updated this and few other instances (missed that some time ago clearly)

PR #321, opened 2023-12-06
Initial implementation of VercelKV apl

lkostrowski:
Docs will follow when tested and moved out of experimental

### DIFF
--- a/src/APL/vercel-kv/vercel-kv-apl.ts
+++ b/src/APL/vercel-kv/vercel-kv-apl.ts
@@ -1,4 +1,4 @@
-import { kv, VercelKV } from "@vercel/kv";
+import { kv } from "@vercel/kv";
 
 import { APL, AplConfiguredResult, AplReadyResult, AuthData } from "../apl";
 import { createAPLDebug } from "../apl-debug";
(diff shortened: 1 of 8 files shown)


src/APL/vercel-kv/vercel-kv-apl.ts
```diff
@@ -0,0 +1,123 @@
...
+
+    if (results === null) {
+      throw new Error("Missing KV collection, data was never written");
+    }
+
+    return Object.values(results).reduce((collectionOfAuthData, item) => {
+      const authData = JSON.parse(item) as AuthData;
+
+      return [...collectionOfAuthData, authData];
+    }, [] as AuthData[]);
```
witoszekdev, line 83: nitpick: Do we need to use reduce here? I think a `map` would be more readable:

```suggestion
    return Object.values(results).map((item) => {
      const authData = JSON.parse(item) as AuthData;
      return authData;
    });
```

src/APL/vercel-kv/vercel-kv-apl.ts
```diff
@@ -0,0 +1,123 @@
...
+    if (!this.envVariablesRequiredByKvExist()) {
+      return {
+        configured: false,
+        error: new Error("Missing KV env variables, please link KV storage to your project"),
+      };
+    }
+
+    return {
+      configured: true,
+    };
```
witoszekdev, line 109: nitpick: let's make this the same way as `isReady`

```suggestion
    const configured = this.envVariablesRequiredByKvExist();
    
    return configured ? {
      configured: true
    } : {
      configured: false,
      error: new Error("Missing KV env variables, please link KV storage to your project")
    };
```

witoszekdev (approved):

src/APL/vercel-kv/vercel-kv-apl.ts
lkostrowski: yes, I started with object and then forgot to change it 👍

PR #335, opened 2024-01-04
Add otel for vercel kv apl

lkostrowski:


src/APL/vercel-kv/vercel-kv-apl.ts
```diff
@@ -29,44 +34,133 @@ export class VercelKvApl implements APL {
...
+              message: "Received response from VercelKV",
+            })
+            .end();
+
+          return authData ?? undefined;
+        } catch (e) {
+          this.debug("Failed to get auth data from Vercel KV");
+          this.debug(e);
+
+          span.recordException("Failed to get auth data from Vercel KV");
```
witoszekdev, line 69: question: how is this different than `span.setStatus`?

src/APL/vercel-kv/vercel-kv-apl.ts
```diff
@@ -29,44 +34,133 @@ export class VercelKvApl implements APL {
...
-      this.debug("Failed to set auth data in Vercel KV");
-      this.debug(e);
-
-      throw e;
-    }
+    return this.tracer.startActiveSpan(
+      "VercelKvApl.set",
+      {
+        attributes: {
+          saleorApiUrl: authData.saleorApiUrl,
```
witoszekdev, line 91:
```suggestion
          saleorApiUrl: authData.saleorApiUrl,
          appId: authData.appId,
```

I think we should add `appId`, the same way as in Cloud APL

witoszekdev (changes requested):

src/APL/vercel-kv/vercel-kv-apl.ts
lkostrowski: AFAIK:
- setStatus marks span to be error, its long-term operation with a result
- recordException is a single point of time, like log. It sets precise event on span, that it happened just there. 

I'm not sure if this is correctly displayed in Datadog, but since we are still exploring this, I'd rather leave it and have more information. App-sdk itself doesn't send otel logs, so this could be a way to have them

witoszekdev (approved):

PR #338, opened 2024-02-13
Use history API to notify about routing updates

typeofweb:
Fixes DEL-67

Related: https://github.com/saleor/app-sdk/pull/338


mmiszy:
I've left it as a draft because we need to verify if any additional methods need to be patched as well.

lkostrowski:
stale, closing

PR #339, opened 2024-02-20
Parse saleor schema version and add it to context

krzysztofzuraw:
Parse the `saleor-schema-version` header and include the parsed version in the contexts of `createHandler` and `webhookFactory`. If the header is absent (Saleor version below 3.15), the version will default to 0. 

Connected [RFC](https://github.com/saleor/apps/issues/1213).

### DIFF
--- a/src/handlers/next/create-manifest-handler.ts
+++ b/src/handlers/next/create-manifest-handler.ts
@@ -1,12 +1,13 @@
 import { NextApiHandler, NextApiRequest } from "next";
 
-import { getBaseUrl } from "../../headers";
+import { getBaseUrl, getSaleorHeaders } from "../../headers";
 import { AppManifest } from "../../types";
 
 export type CreateManifestHandlerOptions = {
   manifestFactory(context: {
     appBaseUrl: string;
     request: NextApiRequest;
+    schemaVersion: number;
   }): AppManifest | Promise<AppManifest>;
 };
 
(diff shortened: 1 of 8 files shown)


src/handlers/next/create-manifest-handler.ts
```diff
@@ -1,12 +1,13 @@
...
 
-import { getBaseUrl } from "../../headers";
+import { getBaseUrl, getSaleorHeaders } from "../../headers";
 import { AppManifest } from "../../types";
 
 export type CreateManifestHandlerOptions = {
   manifestFactory(context: {
     appBaseUrl: string;
     request: NextApiRequest;
+    schemaVersion: number;
```
lkostrowski, line 10: I dont think it should fallback to zero. I think it should be `number | null`. If field is not supported, then its not supported. No need for ambiguous values

src/handlers/next/saleor-webhooks/process-saleor-webhook.test.ts
```diff
@@ -149,4 +150,27 @@ describe("processAsyncSaleorWebhook", () => {
       })
     ).rejects.toThrow("Request signature check failed");
   });
+
+  it("Fallback to 0 if saleor-schema-version header is missing", async () => {
```
lkostrowski, line 154: please add test for parsing correct value

lkostrowski (changes requested):

src/handlers/next/create-manifest-handler.ts
```diff
@@ -1,12 +1,13 @@
...
 
-import { getBaseUrl } from "../../headers";
+import { getBaseUrl, getSaleorHeaders } from "../../headers";
 import { AppManifest } from "../../types";
 
 export type CreateManifestHandlerOptions = {
   manifestFactory(context: {
     appBaseUrl: string;
     request: NextApiRequest;
+    schemaVersion: number | null;
```
lkostrowski, line 10: Non blocking; you can add comment (jsdoc) mentioning here that null means <3.15 

IDEs are using comments for hints

lkostrowski (approved):

PR #341, opened 2024-02-28
Parse schema version when processing saleor webhook

krzysztofzuraw:
Fix wrong logic introduced in [0.49.0](https://github.com/saleor/app-sdk/releases/tag/v0.49.0): there is not header `saleor-schema-version` when app-sdk is processing saleor webhook. This header is only present on install request.

Now app-sdk will try to parse version from `version` field on GraphQL subscription [Event](https://docs.saleor.io/docs/3.x/api-storefront/miscellaneous/interfaces/event#code-style-fontweight-normal-eventbversionbcodestring-). If field is not present `null` will be returned.

### DIFF
--- a/src/handlers/next/saleor-webhooks/process-saleor-webhook.test.ts
+++ b/src/handlers/next/saleor-webhooks/process-saleor-webhook.test.ts
@@ -172,26 +172,4 @@ describe("processAsyncSaleorWebhook", () => {
       schemaVersion: null,
     });
   });
-
-  it("Return schema version if saleor-schema-version header is present", async () => {
-    await expect(
-      processSaleorWebhook({
-        req: mockRequest,
-        apl: mockAPL,
-        allowedEvent: "PRODUCT_UPDATED",
-      })
-    ).resolves.toStrictEqual({
-      authData: {
-        appId: "mock-app-id",
-        domain: "example.com",
-        jwks: "{}",
-        saleorApiUrl: "https://example.com/graphql/",
-        token: "mock-token",
-      },
-      baseUrl: "https://some-saleor-host.cloud",
-      event: "product_updated",
-      payload: {},
-      schemaVersion: null,
-    });
-  });
 });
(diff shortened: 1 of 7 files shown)


src/handlers/next/saleor-webhooks/process-saleor-webhook.test.ts
```diff
@@ -173,26 +172,4 @@ describe("processAsyncSaleorWebhook", () => {
       schemaVersion: null,
     });
   });
-
-  it("Return schema version if saleor-schema-version header is present", async () => {
```
krzysztofzuraw, line 177: I had a problem with this test - currently `version` will be on payload but we mock full `raw-body` in the top of the tests. I tried:
1. Removing mock and passing values directly - I've got a timeouts
2. Using technique from https://github.com/vitest-dev/vitest/discussions/4124 to dynamically asssign value of the mock per test

All without luck - do you have clue what can be done here?

lkostrowski: i dont understand. header will be gone, so this test should be removed. 
you should have test that mocks graphql body and checks if it extracts it, right?

src/handlers/next/saleor-webhooks/process-saleor-webhook.ts
```diff
@@ -140,7 +137,7 @@ export const processSaleorWebhook: ProcessSaleorWebhook = async <T>({
           throw new WebhookError("Missing request body", "MISSING_REQUEST_BODY");
         }
 
-        let parsedBody: unknown;
+        let parsedBody: unknown & { version?: string | null };
```
lkostrowski, line 140: maybe `schemaVersion`? "version" is too generic

lkostrowski (changes requested):

src/handlers/next/saleor-webhooks/process-saleor-webhook.ts
krzysztofzuraw: field on `Event` is named `version`: https://docs.saleor.io/docs/3.x/api-storefront/miscellaneous/interfaces/event#code-style-fontweight-normal-eventbversionbcodestring-

lkostrowski (approved):

PR #344, opened 2024-02-29
Change CI/CD workflows

kzuraw:
I changed a bit CI/CD workflows:
1. Releasing dev tag on npm will be available after you add `release dev tag` label
2. I simplified `main` workflow to have one action for lint and tests (we will need to update required checks on branch)
3. Release workflow now uses PAT token

Ref SHOPX-315

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,4 +1,3 @@
-
 version: 2
 updates:
   - package-ecosystem: "npm"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ jobs:
         with:
           run_install: |
             - args: [--frozen-lockfile]
-      - name: Check linters
+      - name: Check linter
         run: pnpm lint
       - name: Check types
         run: pnpm check-types
--- a/src/APL/upstash-apl.test.ts
+++ b/src/APL/upstash-apl.test.ts
@@ -34,10 +34,10 @@ describe("APL", () => {
       process.env[UpstashAPLVariables.UPSTASH_URL] = "environment";
 
       const apl = await new UpstashAPL({ restToken: "option", restURL: "option" });
-      // eslint-disable-next-line dot-notation
-      expect(apl["restToken"]).toBe("option");
-      // eslint-disable-next-line dot-notation
-      expect(apl["restURL"]).toBe("option");
+      // @ts-expect-error - testing private properties
+      expect(apl.restToken).toBe("option");
+      // @ts-expect-error - testing private properties
+      expect(apl.restURL).toBe("option");
     });
 
     describe("set", () => {
--- a/src/util/extract-app-permissions-from-jwt.test.ts
+++ b/src/util/extract-app-permissions-from-jwt.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect,it } from "vitest";
+import { describe, expect, it } from "vitest";
 
 import { extractAppPermissionsFromJwt } from "./extract-app-permissions-from-jwt";
 


.github/workflows/dev-release.yml
```diff
@@ -0,0 +1,48 @@
+name: Release @dev tag to npm
+on:
+  pull_request:
+    types: [labeled]
```
lkostrowski, line 4: It will trigger on label, right? So if I want to re-create the release, I need to remove the label and add it again, OR add any extra label after the "release dev tag" label?

I think its fine for starters, but we should likely replace it with "comment" event, to have more control ;)

.github/workflows/dev-release.yml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+  release:
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'release dev tag') }}
```
lkostrowski, line 8: maybe you can check with label was ADDED in this run (payload) instead what are all labels?

so 
1. "foobar" added → do nothing
2. "release dev tag" added → release
3. "barbaz" added → **do nothing, but currently it will release again**

.github/workflows/dev-release.yml
```diff
@@ -0,0 +1,48 @@
...
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
```
lkostrowski, line 11: add hash

.github/workflows/dev-release.yml
```diff
@@ -0,0 +1,48 @@
...
+        with:
+          fetch-depth: 0
+      - name: Setup PNPM
+        uses: pnpm/action-setup@a3252b78c470c02df07e9d59298aecedc3ccdd6d # v3.0.0
+        with:
+          run_install: |
+            - args: [--frozen-lockfile]
+      - name: Check for changeset
+        run: pnpm exec changeset status --since origin/main
```
lkostrowski, line 20: is this doing anything except error?
https://github.com/changesets/changesets/blob/main/docs/command-line-options.md#status

If no changeset is there, it will kill the workflow. Should it? Maybe it should gracefully exit?

lkostrowski (changes requested):

.github/workflows/dev-release.yml
krzysztofzuraw: I just tested: 
1. if you have `release dev tag` and you add new label - workflow will be triggered (as you still have trigger label label)
2. if you want trigger release again - remove and add label again

.github/workflows/dev-release.yml
krzysztofzuraw: We hash only 3rd party actions (other than GitHub) per guide:

https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

.github/workflows/dev-release.yml
krzysztofzuraw: Yes - it will fail. Idea here is that should prevent releasing dev tag if you don't have changeset inside your PR as releasing won't work anyway.

lkostrowski: Yea, but it will also send me an email that workflow failed. I think it should be my decision here - I don't want changeset (I edited readme for example), so this PR should just gracefully stop, not fail

krzysztofzuraw: Yes - but it will be your decision to add `label` anyway 😄 . This workflow will only trigger if you add label so I think we need this check in place to avoid situation when you want to release snapshot without changesets 🤔

.github/workflows/dev-release.yml
krzysztofzuraw: Good idea fixed - now we check the last label added in event. It means if you already have `release dev tag` label and add other label workflow will not be triggered.

lkostrowski (approved):

PR #358, opened 2024-03-08
Add support for Fetch API

witoszekdev:
This PR adds experimental support for [Next.js Edge Runtime](https://nextjs.org/docs/pages/api-reference/edge) + other runtimes, which uses Request and Response from Web API


src/fetch-middleware/with-method.ts
```diff
@@ -0,0 +1,25 @@
...
+  DELETE: "DELETE",
+} as const;
+export type HTTPMethod = typeof HTTPMethod[keyof typeof HTTPMethod];
+
+export const withMethod =
+  (...methods: HTTPMethod[]): FetchMiddleware =>
+  (handler) =>
+  async (request) => {
+    if (!methods.includes(request.method as HTTPMethod)) {
+      return new Response("Method not allowed", { status: 405 });
```
mmiszy, line 19: HTTP status 405 _must_ also return `Allow` header with allowed methods

src/fetch-middleware/with-saleor-domain-present.ts
```diff
@@ -0,0 +1,25 @@
+import { getSaleorHeadersFetchAPI } from "../headers";
+import { createFetchMiddlewareDebug } from "./middleware-debug";
+import { FetchMiddleware } from "./types";
+
+const debug = createFetchMiddlewareDebug("withSaleorDomainPresent");
+
+export const withSaleorDomainPresent: FetchMiddleware = (handler) => async (request) => {
```
mmiszy, line 7: Should we keep on using it? I believe `domain` is soft-deprecated and we should switch to `saleorApiUrl` instead

witoszekdev:
Closed in favor of: https://github.com/saleor/app-sdk/pull/380

PR #365, opened 2024-08-27
Prevent save data in apl when data already exists

poulch:


lkostrowski (changes requested):
1. this will prevent *all* installations except first to work, because APL is not pruned after app is uninstalled
2. overwriting APL is something we want
3. users can't install 2 apps with the same id, so this case should not happen (except old installations)

PR #374, opened 2025-01-07
feat: adding redis APL

JannikZed:
this PR is used to bring native redis support as APL. It is fully tested and working.


package.json
```diff
@@ -25,7 +25,8 @@
     "graphql": ">=16.6.0",
     "next": ">=12",
     "react": ">=17",
-    "react-dom": ">=17"
+    "react-dom": ">=17",
+    "redis": "^4.6.12"
```
krzysztofzuraw, line 29: **Suggestion:** what about having redis here pinned as `react` ?

Thanks to that I can use `4.6.x` and `4.7.x` redis as well

package.json
```diff
@@ -169,5 +173,5 @@
     "url": "https://github.com/saleor/saleor-app-sdk/issues"
   },
   "homepage": "https://github.com/saleor/saleor-app-sdk#readme",
-  "packageManager": "pnpm@8.2.0"
+  "packageManager": "pnpm@9.12.0"
```
krzysztofzuraw, line 176: **Question:** can you add this change in other PR?

I think this is good change but I would rather have this in other PR not to have additional changes not related to redis APL.

src/APL/redis/redis-apl.ts
```diff
@@ -0,0 +1,247 @@
...
+      process.env[RedisAPLVariables.REDIS_HASH_KEY] ||
+      "saleor_app_auth";
+
+    if (!redisUrl) {
+      throw new Error(
+        `Missing Redis configuration. Please provide Redis URL either through constructor or ${RedisAPLVariables.REDIS_URL} environment variable.`
+      );
+    }
+
+    this.client = createClient({
```
krzysztofzuraw, line 41: **Question:** what about having client as argument to `constructor` instead?

Thanks to that I can reuse redis client created in my app

src/APL/redis/redis-apl.ts
```diff
@@ -0,0 +1,247 @@
...
+
+  private tracer = getOtelTracer();
+
+  private client: ReturnType<typeof createClient>;
+
+  private hashCollectionKey: string;
+
+  constructor(config?: RedisAPLConfig) {
+    const redisUrl = config?.url || process.env[RedisAPLVariables.REDIS_URL];
+    this.hashCollectionKey =
```
krzysztofzuraw, line 30: **Suggestion:** I would move this after check if have have `redisUrl` so we exit early

package.json
JannikZed: got it! I just had to do it, as I could no longer use pnpm 8.2.0 because:  `GET https://registry.npmjs.org/clean-publish error (ERR_INVALID_THIS). Will retry in 10 seconds. 2 retries left.`
But I removed it for this PR

src/APL/redis/redis-apl.ts
JannikZed: sure, we can do it in that way.

package.json
JannikZed: done

src/APL/redis/redis-apl.ts
JannikZed: when we get the client as argument we no longer need this check.

src/APL/redis/redis-apl.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  private client: ReturnType<typeof createClient>;
+
+  private hashCollectionKey: string;
+
+  constructor(config: RedisAPLConfig) {
+    this.client = config.client;
+    this.hashCollectionKey =
+      config.hashCollectionKey ||
+      process.env[RedisAPLVariables.REDIS_HASH_KEY] ||
```
krzysztofzuraw, line 63: **Question:** do you need this env variable?

I think you can do (in your app code):

```
const apl = new RedisAPL({
   client, 
   hashCollectionKey: process.env.REDIS_HASH_KEY
  });
```

src/APL/redis/redis-apl.ts
```diff
@@ -0,0 +1,267 @@
...
+ * @example
+ * ```typescript
+ * // Create and configure Redis client
+ * const client = createClient({
+ *   url: "redis://localhost:6379",
+ *   // Add any additional Redis configuration options
+ * });
+ *
+ * // Connect to Redis
+ * await client.connect();
```
krzysztofzuraw, line 36: **Question:** do you need first to connect to Redis? won't `ensureConnection` private function do this for me?

src/APL/redis/redis-apl.ts
```diff
@@ -0,0 +1,267 @@
...
+ *
+ * // Initialize RedisAPL with the client
+ * const apl = new RedisAPL({
+ *   client,
+ *   // Optional: customize the hash collection key
+ *   hashCollectionKey: "my_custom_auth_key"
+ * });
+ *
+ * // Use the APL in your app
+ * await apl.set("app-id", { token: "auth-token" });
```
krzysztofzuraw, line 46: **Suggestion:** for sake of completeness I would improve this example to set full `AuthData`:

```ts
  token: string;
  saleorApiUrl: string;
  appId: string;
```

src/APL/redis/redis-apl.ts
```diff
@@ -0,0 +1,267 @@
...
+ * const authData = await apl.get("app-id");
```
krzysztofzuraw, line 47: **Suggestion:** same here - I think it will help if it will be `apl.get("saleorApiUrl")`

src/APL/redis/redis-apl.test.ts
```diff
@@ -0,0 +1,178 @@
...
+  options: {},
+  duplicate: vi.fn(),
+  multi: vi.fn(),
+  isPubSubActive: false,
+  quit: vi.fn().mockResolvedValue(undefined),
+  sendCommand: vi.fn(),
+  executeIsolated: vi.fn(),
+} as unknown as RedisClientType;
+
+vi.mock("redis", () => ({
```
krzysztofzuraw, line 41: **Question:** do you need to mock full `redis` package? 

I think you already is passing `client` to `RedisAPL`

src/APL/redis/redis-apl.test.ts
```diff
@@ -0,0 +1,178 @@
+import { RedisClientType } from "redis";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { AuthData } from "../apl";
+import { RedisAPL } from "./redis-apl";
+
+// Create a variable to control the connection state
+let isRedisOpen = false;
```
krzysztofzuraw, line 8: **Question:** do you need this variable?

I'm asking as I see it is set only in `beforeEach`

src/APL/redis/redis-apl.test.ts
```diff
@@ -0,0 +1,178 @@
...
+  beforeEach(() => {
+    process.env.REDIS_HASH_KEY = mockHashKey;
+    apl = new RedisAPL({ client: mockRedisClient });
+    vi.clearAllMocks();
+    isRedisOpen = false;
+  });
+
+  afterEach(async () => {
+    delete process.env.REDIS_HASH_KEY;
+    if (mockRedisClient.isOpen) {
```
krzysztofzuraw, line 64: **Question:** I assume that this mock - do we need to disconnect?

src/APL/redis/redis-apl.ts
```diff
@@ -0,0 +1,267 @@
...
+import { getOtelTracer, OTEL_APL_SERVICE_NAME } from "../../open-telemetry";
+import { APL, AplConfiguredResult, AplReadyResult, AuthData } from "../apl";
+import { createAPLDebug } from "../apl-debug";
+
+/**
+ * Configuration options for RedisAPL
+ */
+type RedisAPLConfig = {
+  /** Redis client instance to use for storage */
+  client: ReturnType<typeof createClient>;
```
lkostrowski, line 14: you only use / need some methods of the client (hGet, hSet etc), so you can do

```
type Client = ReturnType<typeof createClient>

client: Pick<Client, 'hGet' | 'hSet' | ...>

```

now, testing is easier and only these methods must be implemented / mocked. Class should not request methods it doesnt need

src/APL/redis/redis-apl.ts
lkostrowski: Yup, better to move this to the caller side

lkostrowski (changes requested):

krzysztofzuraw:
@JannikZed please update your branch - it should now have pnpm 9.12.3 :)

src/APL/redis/redis-apl.test.ts
```diff
@@ -0,0 +1,178 @@
...
+  hDel: mockHDel,
+  hGetAll: mockHGetAll,
+  get isOpen() {
+    return isRedisOpen;
+  },
+  disconnect: mockDisconnect,
+  isReady: false,
+  commandOptions: vi.fn(),
+  options: {},
+  duplicate: vi.fn(),
```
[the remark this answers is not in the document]
JannikZed: yeah that we copied over.. I did make a cleanup now.

src/APL/redis/redis-apl.test.ts
JannikZed: I improved this now and the mocked connect and disconnect are setting the isRedisOpen state now to simulate redis better.

src/APL/redis/redis-apl.test.ts
JannikZed: I improved this now and the mocked connect and disconnect are setting the isRedisOpen state now to simulate redis better.

src/APL/redis/redis-apl.ts
JannikZed: true that.. Sorry that was auto-generated. Fixing now

src/APL/redis/redis-apl.ts
JannikZed: absolutely right. Forgot to remove that again. Fixed now.

src/APL/redis/redis-apl.ts
JannikZed: that is already done.

package.json
```diff
@@ -25,7 +25,8 @@
     "graphql": ">=16.6.0",
     "next": ">=12",
     "react": ">=17",
-    "react-dom": ">=17"
+    "react-dom": ">=17",
+    "redis": ">=4"
```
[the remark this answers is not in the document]
JannikZed: done

src/APL/redis/redis-apl.ts
JannikZed: is that really good practice? The class gets the full redis client and not just certain methods from there. So I don't see that honestly. I tried it out as well and did bring some more typing incompatibilities with it, so I would like to leave that away if it is not important to you.

krzysztofzuraw: I think this is for easier testing / type checking - we need only subset of methods from Redis client - only those we need. Also we are requesting only what we actually use.

I just checked and with:

```diff
diff --git a/src/APL/redis/redis-apl.ts b/src/APL/redis/redis-apl.ts
index d6d8e7e..2867620 100644
--- a/src/APL/redis/redis-apl.ts
+++ b/src/APL/redis/redis-apl.ts
@@ -6,12 +6,17 @@ import { getOtelTracer, OTEL_APL_SERVICE_NAME } from "../../open-telemetry";
 import { APL, AplConfiguredResult, AplReadyResult, AuthData } from "../apl";
 import { createAPLDebug } from "../apl-debug";
 
+type RedisClient = Pick<
+  ReturnType<typeof createClient>,
+  "connect" | "isOpen" | "hGet" | "hSet" | "hDel" | "hGetAll" | "ping"
+>;
+
 /**
  * Configuration options for RedisAPL
  */
 type RedisAPLConfig = {
   /** Redis client instance to use for storage */
-  client: ReturnType<typeof createClient>;
+  client: RedisClient;
   /** Optional key to use for the hash collection. Defaults to "saleor_app_auth" */
   hashCollectionKey?: string;
:
```

works without a problem 🤔 then I can do in test:

```diff
diff --git a/src/APL/redis/redis-apl.test.ts b/src/APL/redis/redis-apl.test.ts
index 855345d..8575982 100644
--- a/src/APL/redis/redis-apl.test.ts
+++ b/src/APL/redis/redis-apl.test.ts
@@ -1,4 +1,4 @@
-import { createClient, RedisClientType } from "redis";
+import { createClient } from "redis";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import { AuthData } from "../apl";
@@ -32,7 +32,7 @@ const mockRedisClient = {
   },
   disconnect: mockDisconnect,
   isReady: false,
-} as unknown as RedisClientType;
+};
 
 vi.mock("redis", () => ({
   createClient: vi.fn(() => mockRedisClient),
```

JannikZed: got it. Changed it as you wish.

krzysztofzuraw (approved):

src/APL/index.ts
```diff
@@ -1,5 +1,6 @@
 export * from "./apl";
 export * from "./env-apl";
 export * from "./file-apl";
+export * from "./redis";
```
lkostrowski, line 4: @JannikZed let's avoid that, now importing any other APL will evaluate redis, this causes issues, e.g. next is poor at tree shaking, possibly fail because it can't import redis sdk

Please check package.json and add `exports` key with new path, like vercel-kv. For example 

```
    "./APL/redis": {
      "types": "./APL/redis/index.d.ts",
      "import": "./APL/redis/index.mjs",
      "require": "./APL/redis/index.js"
    },
```

also new entry point will be needed

https://github.com/saleor/app-sdk/blob/main/tsup.config.ts#L13

lkostrowski (changes requested):

src/APL/index.ts
JannikZed: @IKarbowiak alright, check it out, if that works for you.
Is there a very easy way to test also if the package got build correctly and we can actually use the different APLs? are you spinning up an example saleor app and installing it locally or how are you doing that?

JannikZed:
@krzysztofzuraw @lkostrowski I added redis integration tests now as well. They don't run automatically, as they need a running redis instance. I think it can help a lot to keep the quality high, as we test if the build is working and the final package can actually work with a real redis. Let me know, what you think. I can quickly remove it again, if that is not what you are expecting.

lkostrowski:
@JannikZed I'm fine with integration, but it has to be at least in readme how to run it. Best if you prepare docker-compose and document how to run it from scratch (run docker container, then run tests etc)

JannikZed:
@lkostrowski got it - included that now

lkostrowski (approved):

JannikZed:
@andrzejewsky so it is just you remaining :)

PR #380, opened 2025-01-14
New handlers: refactored, Web API and AWS Lambda support

witoszekdev:
## Description

This PR introduces new handlers for Fetch Web API and AWS Lambda and refactors existing code so that it can easily allow adding new platforms in the future.

After the change handlers code is splits them into 3 parts:

1. **Platform Adapter** - class that is responsible of:
	- parsing received requests from specific platform format (e.g. Next.js request, Web API Request) into a common format
	- and preparing response from common format into platform specific format (e.g. Next.js response, Web API Response)
2. **Handler use cases** - class that represents business logic specific to handler, has a single method that returns a result
	- e.g. register use case: readers all required headers from request, validates it, saves to apl, returns success or failure
3. **Handler factory method** - function that creates a handler instance with all dependencies injected, each platform has separate folder, for example:
	- `src/handlers/nextjs` - For handlers that use Next.js platform adapter
	- `src/handlers/fetch-api` - For handlers that use Fetch API platform adapter

This way we'll be able to easily add new platforms and handlers without changing existing code.

Currently in this PR:
- Next.js - existing implementation, shouldn't be changed
- Web API - new implementation
- AWS Lambda event - new implementation

## Examples and docs

**Usage examples**:

https://github.com/witoszekdev/saleor-app-hono-aws-lambda-template
https://github.com/witoszekdev/saleor-app-hono-deno-template
https://github.com/witoszekdev/saleor-app-hono-cf-pages-template
https://github.com/saleor/saleor-app-template/pull/267

**Docs**:

https://github.com/saleor/saleor-docs/pull/1437

## Breaking changes?

`SaleorWebhook` accepts `onError` and `formatErrorResponse` parameters callbacks, now they won't pass `res` object in Next.js runtime, since these methods **weren't supposed to send responses**


src/handlers/shared/generic-adapter-use-case-types.ts
```diff
@@ -0,0 +1,52 @@
...
+  | {
+      status: ResultStatusCodes;
+      body: Body;
+      bodyType: "json";
+    }
+  | {
+      status: ResultStatusCodes;
+      body: string;
+      bodyType: "string";
+    };
```
lkostrowski, line 32: we can create new Response() here and use it as common, standard interface
then platform-specific handler can convert it to non-response (or just return it)

src/handlers/shared/generic-adapter-use-case-types.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Interface for adapters that translate specific platform objects (e.g. Web API, Next.js)
+ * into a common interface that can be used in each handler use case
+ * */
+export interface PlatformAdapterInterface<T extends HandlerInput = HandlerInput> {
+  send(result: HandlerUseCaseResult): unknown;
```
lkostrowski, line 39: consider making the interface thin. You only need getHeader and getBody for shared operations, other operations can be non-shared, hence better flexibility

src/handlers/shared/generic-adapter-use-case-types.ts
```diff
@@ -0,0 +1,52 @@
...
+  request: T;
+}
+
+/** Interfaces for use case handlers that encapsulate business logic
+ * (e.g. validating headers, checking HTTP method, etc. ) */
+export interface HandlerUseCaseInterface {
+  getResult(): Promise<HandlerUseCaseResult>;
+}
+
+export type HandlerInput = NextJsHandlerInput | WebApiHandlerInput | AwsLambdaHandlerInput;
```
lkostrowski, line 52: its a cricular dep, we should avoid this. you can make generic type plain object and allow users (write in docs) to fill the generic with speciifc platform tpye

src/handlers/shared/generic-adapter-use-case-types.ts
lkostrowski: if you extract them from the interface you can rename "send" method

src/handlers/shared/register-use-case.ts
```diff
@@ -0,0 +1,356 @@
...
+    adapter,
+    config,
+  }: {
+    adapter: PlatformAdapterInterface<I>;
+    config: GenericCreateAppRegisterHandlerOptions<I>;
+  }) {
+    this.adapter = adapter;
+    this.adapterMiddleware = new PlatformAdapterMiddleware(adapter);
+    this.config = config;
+  }
```
lkostrowski, line 155:
```class XX {
  private adapter = new PlatformAdapterMiddleware(this.params.adapter);

constructor (private params: {...})
}```

suggestion

File: src/handlers/shared/register-use-case.ts [hunk 5]
```diff
@@ -0,0 +1,356 @@
...
+    return null;
+  }
+
+  /** TODO: Add missing retes methods:
+   * withMethod("POST"),
+   * withSaleorDomainPresent,
+   * withAuthTokenRequired,
+   * baseHandler,
+   * */
+  async getResult(): Promise<HandlerUseCaseResult> {
```
lkostrowski, line 178: consider moving `config` to getResult mehtod

src/handlers/shared/register-use-case.ts
```diff
@@ -0,0 +1,356 @@
...
+    for (const check of checksToRun) {
+      if (check) {
+        return check;
+      }
+    }
+
+    return null;
+  }
+
+  /** TODO: Add missing retes methods:
```
lkostrowski, line 172: to remove

src/handlers/shared/register-use-case.ts
```diff
@@ -0,0 +1,356 @@
...
+
+    const authData = {
+      domain: saleorDomain,
+      token: authToken,
+      saleorApiUrl,
+      appId,
+      jwks,
+    };
+
+    if (this.config.onRequestVerified) {
```
lkostrowski, line 296: consider adding private methods for branches like this for visibility

src/handlers/shared/adapter-middleware.ts
```diff
@@ -0,0 +1,68 @@
...
+  HTTPMethod,
+  PlatformAdapterInterface,
+} from "./generic-adapter-use-case-types";
+
+const debug = createMiddlewareDebug("PlatformAdapterMiddleware");
+
+export class PlatformAdapterMiddleware {
+  constructor(private adapter: PlatformAdapterInterface) {}
+
+  withMethod(methods: HTTPMethod[]): HandlerUseCaseResult | null {
```
lkostrowski, line 21: validateMethod?

src/handlers/shared/adapter-middleware.ts
```diff
@@ -0,0 +1,68 @@
...
+    if (!methods.includes(this.adapter.method)) {
+      return {
+        body: "Method not allowed",
+        bodyType: "string",
+        status: 405,
+      };
+    }
+
+    return null;
```
lkostrowski, line 30: maybe just dont return anything

src/handlers/shared/adapter-middleware.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  withSaleorDomainPresent(): HandlerUseCaseResult | null {
```
lkostrowski, line 33: create a todo to check saleorApiUrl. Domain is legacy, we should slowly remove it and check against saleorApiUrl

do not do it in this pr to avoid changes within refactor code

src/handlers/next/create-app-register-handler.ts
```diff
@@ -49,197 +16,15 @@ export const createRegisterHandlerResponseBody = (
...
-      );
-    }
-
-    debug("Register  complete");
-
-    return Response.OK(createRegisterHandlerResponseBody(true));
+export type CreateAppRegisterHandlerOptions =
+  GenericCreateAppRegisterHandlerOptions<NextJsHandlerInput>;
+
+export const createAppRegisterHandler =
```
lkostrowski, line 23: add JSDoc comments - describe the platform, e.g this is a Pages Router

src/handlers/fetch-api/platform-adapter.ts
```diff
@@ -0,0 +1,39 @@
+import {
+  HandlerUseCaseResult,
+  PlatformAdapterInterface,
+} from "../shared/generic-adapter-use-case-types";
+
+export type WebApiHandlerInput = Request;
+export type WebApiHandler = (req: Request) => Response | Promise<Response>;
+
+export class WebApiAdapter implements PlatformAdapterInterface<WebApiHandlerInput> {
```
lkostrowski, line 9: add alias (in package json) to provide a separate path like `handlers/next-app-router` for better DX

package.json
```diff
@@ -21,7 +21,9 @@
   "author": "",
   "license": "ISC",
   "peerDependencies": {
+    "@types/aws-lambda": "8.10.147",
```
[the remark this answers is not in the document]
witoszekdev: I moved it to `devDependency` since these are just types, no runtime code involved (unlike vercel/kv or aws-lambda - not types)

.changeset/fluffy-cheetahs-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+"@saleor/app-sdk": minor
```
lkostrowski, line 2: please make it major since it has breaking changes

.changeset/fluffy-cheetahs-sleep.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Added `handlers/fetch-api` which adds support for frameworks that use [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)
```
lkostrowski, line 5: please remember to add full changeset and docs

package.json
```diff
@@ -133,6 +145,26 @@
...
+    "./handlers/fetch-api": {
+      "types": "./handlers/fetch-api/index.d.ts",
+      "import": "./handlers/fetch-api/index.mjs",
+      "require": "./handlers/fetch-api/index.js"
+    },
+    "./handlers/next-app-router": {
+      "types": "./handlers/fetch-api/index.d.ts",
+      "import": "./handlers/fetch-api/index.mjs",
+      "require": "./handlers/fetch-api/index.js"
+    },
```
lkostrowski, line 157: won't it have the same names as fetch api? It would be confusing, probably better to also add alias on export

so `next-app-router` file can include `export { FetchApi as  NextAppRouterWebhook } from '../fetch-api'` or something like this

lkostrowski: would be easier if we see docs at this point

lkostrowski (changes requested):
It would be easier for me to review if I already see docs with examples. In case of SDK public API is more important than the code

lkostrowski (commented):
another partial review ;)

witoszekdev:
@lkostrowski Public API usage doesn't change, the change made to existing Next.js handlers are meant to be drop-in and non-breaking. New handlers for Web API and Lambda have the same signature as Next.js handlers, they only differ when accepting a `request` object.

I've prepared examples here:
- https://github.com/witoszekdev/saleor-app-hono-aws-lambda-template
- https://github.com/witoszekdev/saleor-app-hono-deno-template
- https://github.com/witoszekdev/saleor-app-hono-cf-pages-template
- https://github.com/saleor/saleor-app-template/pull/267

And docs here: https://github.com/saleor/saleor-docs/pull/1437

.changeset/fluffy-cheetahs-sleep.md
witoszekdev: https://github.com/saleor/saleor-docs/pull/1437

src/handlers/actions/manifest-action-handler.ts
```diff
@@ -0,0 +1,57 @@
...
+        schemaVersion,
+      });
+
+      debug("Executed manifest file");
+
+      return {
+        status: 200,
+        bodyType: "json",
+        body: manifest,
+      };
```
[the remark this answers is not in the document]
witoszekdev: I thought that using our own interface for responses would be better than using native Response, for the following reasons:
- we don't know what the underling platform code will be used, it might as well be grpc, this abstraction is more high-level than `Response` object
- if we would use `Response` object, then there's no clear indication if the data is JSON or plain text since it stores the response data as `ReadableStream` and data type as `content-type` header - I think it would be more complicated to read the stream, check header, than simply pass an object
- creating `Response` object and then reading the data again as `json` or `text` adds additional overhead (we need to create stream, then read it, and parse the json VS pass the object directly)

witoszekdev:
@lkostrowski one more thing: `fetch-middleware` was a leftover from old PR that I resumed work on and I actually didn't intend on including these changes.

I'm also not sure if we even want to add these middlewares to begin with? They are not used in other places in app-sdk. Previously we had them because they were passed to retes.

I also don't think these are used in any of our apps, since most of them use higher level abstractions by using `/handlers`. I think this is a better approach, since using low-level middlewares could expose users of app-sdk to malicious requests if not all validations are used in their application code. I think it's much easier to rely on a handler that does all the validation for you.

src/handlers/actions/register-action-handler.ts
```diff
@@ -0,0 +1,506 @@
...
+export class RegisterActionHandler<I> implements ActionHandlerInterface {
+  constructor(private adapter: PlatformAdapterInterface<I>) {}
+
+  private adapterMiddleware = new PlatformAdapterMiddleware(this.adapter);
+
+  private runPreChecks(): ActionHandlerResult | null {
+    const checksToRun = [
+      this.adapterMiddleware.withMethod(["POST"]),
+      this.adapterMiddleware.withSaleorDomainPresent(),
+    ];
```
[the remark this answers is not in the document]
witoszekdev: I don't think modifying `Request` object is a good idea. It would be like adding new properties to globals like `Array` or `Window` 😬 These objects are meant to be immutable.

Adding context is usually done by web frameworks that use Fetch API by creating new Context object and passing it everywhere, or by using AsyncLocalStorage. [Example in Hono](https://hono.dev/docs/api/context).

We could do this, but then we would need to have some context mechanism of our own. We could also use AsyncLocalStorage, but this is also problematic since runtimes like Cloudflare Workers, or Deno require enabling Node compatibility layer to do this :/

src/handlers/actions/register-action-handler.ts
```diff
@@ -0,0 +1,506 @@
...
+      saleorDomain,
+    });
+
+    if (handleOnRequestResult) {
+      return handleOnRequestResult;
+    }
+
+    if (!saleorApiUrl) {
+      // TODO: We should strictly require `saleorApiUrl` instead of
+      // relying on `saleor-domain` that is deprecated
```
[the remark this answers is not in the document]
witoszekdev: why? the comment is still valid because we are not doing strict check on `saleorApiUrl` for compatibility reasons (I didn't change bussiness logic, rather I just copied our old code and adapted it to work with our new adapter abstraction).

We can make another breaking change and just drop the `Saleor-Domain` usage altogether imho

src/handlers/actions/register-action-handler.ts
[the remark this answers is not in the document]
lkostrowski: Modifying request isn't anything new. It's not about modifying global, but instance. It's not about `Array.prototype.foo` by `(new Array()).foo`

Context you mentioned is a platform context.
What I mean is request context. It can be achieved several ways, eg with async_hooks, modifying request or adding another argument (like `createWebhookHandler(req, res, ctx)`). Or how TRPC middleware works.

You can also read eg https://expressjs.com/en/guide/using-middleware.html

What I want to achieve is to allow middleware to attach data for next middlewares or handlers and compose the rich request in the end.

For example, if your handler fetches AuthData, it can attach it to request (pass to next step), instead of fetching it again in next middleware/handler

src/handlers/actions/manifest-action-handler.ts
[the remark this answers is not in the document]
lkostrowski: fine 👍

package.json
witoszekdev: This is fine, since exported names do not reference Fetch API in any way, names are the same as in Next.js handlers except they are in a different folder. Our exports are:
- `createAppRegisterHandler`
- `createManifestHandler`
- `createProtectedHandler`
- `SaleorSyncWebhook`
- `SaleorAsyncWebhook`

src/handlers/actions/register-action-handler.ts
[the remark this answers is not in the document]
witoszekdev: OK, I think we are talking about different things here and maybe the name middleware in this place is wrong and confusing 😅

These "middlewares" do not have any data we would want to pass as context, since they don't do any asynchronous actions.

This specific register action has a concept of "context" in its config, where you can attach to different events, e.g.:

```ts
  /**
   * Run right after Saleor calls this endpoint
   */
  onRequestStart?(
    request: RequestType,
    context: {
      authToken?: string;
      saleorDomain?: string;
      saleorApiUrl?: string;
      respondWithError: CallbackErrorHandler;
    }
  ): Promise<void>;
```

_Config shape is the same as previous Next.js actions in app-sdk, it just has different `Request` object type depending on platform_

`createProtectedHandler` also has its own context: 

```ts
export type ProtectedHandlerContext = {
  baseUrl: string;
  authData: AuthData;
  user: TokenUserPayload;
};
```

Webhook handlers also have context (no change from previous public API):

```ts
export type SaleorWebhookHandler<TPayload = unknown, TExtras = {}> = (
  req: Request,
  ctx: WebhookContext<TPayload> & TExtras
) => Response | Promise<Response>;
```

As for `fetch-middleware`, please see my comment: https://github.com/saleor/app-sdk/pull/380#issuecomment-2615426783.

witoszekdev:
All features are now merged to `main` branch

PR #396, opened 2025-01-28
Refactor handlers folder structure

witoszekdev:
- Separated shared types to separate `shared` folder
- Moved platform handlers to `platform` folder, leaving public imports unchanged
- Added re-exports in `platform/next` for backwards compatibility (to be removed in v1)
- Added `@/` alias, updated tests to support it


witoszekdev:
FYI: Codecov is calculated incorrectly, because we haven't added new code, rather moving it to different place ;)

witoszekdev:
Closed, since we don't actually need this changes for v0, we'll have a migration guide prepared

PR #397, opened 2025-01-28
Mark middlewares as depreacted

witoszekdev:
Marked middlewares as deprecated to be removed in v1

### DIFF
--- a/.changeset/witty-carrots-occur.md
+++ b/.changeset/witty-carrots-occur.md
@@ -1,0 +1,5 @@
+---
+"@saleor/app-sdk": patch
+---
+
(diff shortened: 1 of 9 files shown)


.changeset/witty-carrots-occur.md
```diff
@@ -0,0 +1,5 @@
+---
+"@saleor/app-sdk": patch
+---
+
+Marked `/middleware`s as deprecated in favor of `/handlers`
```
lkostrowski, line 5: I dont think we even need that, we can write migration guide. Its fine but not necessary

lkostrowski (approved):

PR #398, opened 2025-01-28
Remove middlewares (for v1)

witoszekdev:
Remove deprecated middlewares from v1

They were marked as deprecated in: https://github.com/saleor/app-sdk/pull/397


witoszekdev:
This change has to be done with a refactor since we actually did use one middleware for create-app-register-handler 🙈

PR #399, opened 2025-01-28
Refactor handlers, remove deprecated exports, v1

witoszekdev:
- Separated shared types to separate `shared` folder
- Moved platform handlers to `platform` folder, leaving public imports unchanged
- Removed re-exports added in https://github.com/saleor/app-sdk/pull/396
- Added `@/` alias, updated tests to support it

> [!NOTE]
> There are multiple files visible as "changed" in `handlers/next` and `handlers/platform/next`, but this is wrong, these files were simply moved to a different folder


package.json
```diff
@@ -144,6 +145,11 @@
       "import": "./handlers/next/index.mjs",
       "require": "./handlers/next/index.js"
     },
+    "./handlers/shared": {
+      "types": "./handlers/shared/index.d.ts",
+      "import": "./handlers/shared/index.mjs",
+      "require": "./handlers/shared/index.js"
+    },
```
lkostrowski, line 152: why do you add this export?

tsup.config.ts
```diff
@@ -1,23 +1,28 @@
...
+    "app-bridge/next/index": "src/app-bridge/next/index.ts",
+    "settings-manager/index": "src/settings-manager/index.ts",
+    "handlers/shared/index": "src/handlers/shared/index.ts",
+
+    // Deprecated
+    "middleware/index": "src/middleware/index.ts",
+
+    // Mapped exports
+    "handlers/next/index": "src/handlers/platforms/next/index.ts",
+  },
```
lkostrowski, line 25: What does mean this api change - from array to object? Should keys include `/index`?

src/handlers/shared/protected-handler.ts
```diff
@@ -0,0 +1,20 @@
...
+  | "MISSING_API_URL_HEADER"
+  | "MISSING_AUTHORIZATION_BEARER_HEADER"
+  | "NOT_REGISTERED"
+  | "JWT_VERIFICATION_FAILED"
+  | "NO_APP_ID";
+
+export const ProtectedHandlerErrorCodeMap: Record<SaleorProtectedHandlerError, number> = {
+  OTHER: 500,
+  MISSING_HOST_HEADER: 400,
+  MISSING_DOMAIN_HEADER: 400,
```
lkostrowski, line 14: we dont need it, we drop domain header check

src/handlers/shared/process-saleor-webhook.ts
```diff
@@ -0,0 +1,38 @@
...
+    Object.setPrototypeOf(this, WebhookError.prototype);
+  }
+}
+
+export type WebhookContext<TPayload> = {
+  baseUrl: string;
+  event: string;
+  payload: TPayload;
+  authData: AuthData;
+  /** For Saleor < 3.15 it will be null. */
```
lkostrowski, line 36: we can drop this comment and add in docs that sdk@1 supports Saleor 3.20+

src/handlers/shared/process-saleor-webhook.ts
lkostrowski, line 1: is file name still valid? it contains random things

lkostrowski (approved):

tsup.config.ts
witoszekdev: Yes, otherwise we woudlnt' be able to do mapped exports in tsup

So for example:
```
[
  "src/headers.ts"
]
```

is the same as:

```
{
  headers: "src/headers.ts"
}
```

src/handlers/shared/process-saleor-webhook.ts
witoszekdev: Nope, this should have been in a single file for all saleor webhook types, I combined them 👍🏻

lkostrowski (approved):

PR #401, opened 2025-01-28
Add new Adapter and Action interfaces, add new Next.js handler implementation, remove middlewares

witoszekdev:
This PR adds new interfaces for writing handlers:
- Adapter
- Action
- Validator

It replaces Next.js handlers with these new abstractions.
It removes deprecated process-protected-handler and process-saleor-webhook files, in favor of Validator classes.

It also removes middlewares and `retes` dependency, since they are no longer used in handlers.

Additional tests for existing usage were added: for example, fetching of JWKS when Saleor webhook validation has failed, webhook error handling.

TODO for next PR:
- Add handlers for Web API and AWS lambda
- Make `schemaVersion` required in SaleorWebhook


.changeset/tough-socks-tease.md
```diff
@@ -2,4 +2,6 @@
 "@saleor/app-sdk": major
 ---
 
-Breaking change: Remove checking "domain" header from Saleor requests. It should be replaced with the "saleor-api-url" header.
+Breaking change: Remove checking "saleor-domain" header from Saleor requests. It should be replaced with the "saleor-api-url" header.
```
lkostrowski, line 5: But who should check saleor-api-url? app or sdk? If SDK, it's not a part of the changelog but something we should do

package.json
```diff
@@ -156,6 +150,11 @@
       "import": "./handlers/shared/index.mjs",
       "require": "./handlers/shared/index.js"
     },
+    "./handlers/actions": {
+      "types": "./handlers/actions/index.d.ts",
+      "import": "./handlers/actions/index.mjs",
+      "require": "./handlers/actions/index.js"
+    },
```
lkostrowski, line 157: what is the use case of this being exported? Why App would need that?

src/handlers/actions/manifest-action-handler.ts
```diff
@@ -0,0 +1,71 @@
...
+    const { schemaVersion } = this.adapterMiddleware.getSaleorHeaders();
+    const baseURL = this.adapter.getBaseUrl();
+
+    debug("Received request with schema version \"%s\" and base URL \"%s\"", schemaVersion, baseURL);
+
+    const methodEagerResponse = this.adapterMiddleware.withMethod(["GET"]);
+
+    if (methodEagerResponse) {
+      return methodEagerResponse;
+    }
```
lkostrowski, line 37: not sure if I get that

src/handlers/shared/adapter-middleware.ts
```diff
@@ -0,0 +1,57 @@
...
+  SALEOR_API_URL_HEADER,
+  SALEOR_AUTHORIZATION_BEARER_HEADER,
+  SALEOR_EVENT_HEADER,
+  SALEOR_SCHEMA_VERSION,
+  SALEOR_SIGNATURE_HEADER,
+} from "@/const";
+
+import { HTTPMethod, PlatformAdapterInterface } from "./generic-adapter-use-case-types";
+
+export class PlatformAdapterMiddleware<T> {
```
lkostrowski, line 11: Maybe better name would be to mention what it does (validating requests), not that it uses adapter for it?

src/handlers/shared/protected-action-validator.ts
```diff
@@ -0,0 +1,198 @@
...
+  user: TokenUserPayload;
+};
+
+export type ValidationResult =
+  | { result: "failure"; value: ActionHandlerResult }
+  | { result: "ok"; value: ProtectedHandlerContext };
+
+export class ProtectedActionValidator<I> {
+  // Name left for backwards compatibility
+  private debug = createDebug("processProtectedHandler");
```
lkostrowski, line 30: We dont need this compatibility, we are introdcucing bc

.changeset/tough-socks-tease.md
witoszekdev: Nope, this is something SDK does: we have dropped `saleor-domain` which was deprecated in favor of `saleor-api-url` header in our validation, thus making SDK incompatible with versions prior to 3.15. I've improved the description so that it's more clear.

package.json
lkostrowski: I think on 1.x better to *not* export more than we need. Every time we export something we must be sure it should be there, because its public api and everyone can build on top

Unless someone asks for, I'd prefer to export only platform specific logic, while we control the internals. The we can evolve into exposing this

.changeset/tough-socks-tease.md
```diff
@@ -2,4 +2,6 @@
 "@saleor/app-sdk": major
 ---
 
-Breaking change: Remove checking "domain" header from Saleor requests. It should be replaced with the "saleor-api-url" header.
+Breaking change: SDK will no longer check `saleor-domain` header when validating Saleor requests, instead it will check `saleor-api-url` header.
+
+This makes SDK incompatible with Saleor versions prior to 3.15.
```
lkostrowski, line 7: We don't have to write that, I want SDK 1.x to drop support below 3.20

lkostrowski (approved):

package.json
witoszekdev: All right, I haven't though about that this way 😅 

I'll remove internals export then 👍🏻

PR #404, opened 2025-02-05
Add AWS Lambda handlers

witoszekdev:
This PR adds handlers for AWS Lambda platform for events originating from HTTP events (e.g. Lambda URL, API Gateway).

Example implementation using this PR: https://github.com/witoszekdev/saleor-app-lambda-template

### DIFF
--- a/src/handlers/platforms/aws-lambda/platform-adapter.ts
+++ b/src/handlers/platforms/aws-lambda/platform-adapter.ts
@@ -1,0 +1,105 @@
+import type {
+  APIGatewayProxyEventV2,
+  APIGatewayProxyStructuredResultV2,
+  Context,
+} from "aws-lambda";
+
+import {
+  ActionHandlerResult,
+  HTTPMethod,
+  PlatformAdapterInterface,
+} from "@/handlers/shared/generic-adapter-use-case-types";
+
+export type AwsLambdaHandlerInput = APIGatewayProxyEventV2;
+export type AWSLambdaHandler = (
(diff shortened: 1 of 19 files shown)


lkostrowski (approved):

src/handlers/platforms/aws-lambda/platform-adapter.ts
```diff
@@ -0,0 +1,110 @@
...
+    const name = requestedName.toLocaleLowerCase();
+    return this.request.headers[name] || null;
+  }
+
+  async getBody(): Promise<unknown | null> {
+    if (!this.request.body) {
+      return null;
+    }
+
+    return JSON.parse(this.request.body);
```
poulch, line 52: question: Is it save to run `JSON.parse` without proper error handling in case `this.request.body` will has invalid json?

poulch (approved):

src/handlers/platforms/aws-lambda/platform-adapter.ts
witoszekdev: Yes, we want it to throw error so that we know that request was invalid because of malformated JSON, not that the request was empty (if we would return null).

Ideally I would use [`neverthrow`](https://github.com/supermacro/neverthrow) so that we don't throw exceptions, but I didn't want to introduce another dependency ;)

witoszekdev: In places we use `getBody` we have `try...catch`

PR #444, opened 2025-06-17
Deprecate `buildSyncWebhookResponsePayload`

krzysztofzuraw:
Deprecate `buildSyncWebhookResponsePayload` function. Saleor now exposes JSON schema for webhook response payloads that can be used to generate TypeScript types. See [Saleor docs](https://docs.saleor.io/developer/extending/apps/developing-apps/generating-types-for-sync-webhooks) for more info.

### DIFF
--- a/src/handlers/shared/sync-webhook-response-builder.ts
+++ b/src/handlers/shared/sync-webhook-response-builder.ts
@@ -258,6 +258,8 @@ export type SyncWebhookResponsesMap<V extends SaleorVersion = "3.20"> = CoreSync
 } & TransactionWebhookResponses<V>;
 
 /**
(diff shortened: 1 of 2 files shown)


src/handlers/shared/sync-webhook-response-builder.ts
```diff
@@ -258,6 +258,8 @@ export type SyncWebhookResponsesMap<V extends SaleorVersion = "3.20"> = CoreSync
 } & TransactionWebhookResponses<V>;
 
 /**
+ * @deprecated Saleor now exposes JSON schema for webhook response payloads that can be used to generate TypeScript types. See https://docs.saleor.io/developer/extending/apps/developing-apps/generating-types-for-sync-webhooks for more details.
+ *
```
lkostrowski, line 262: maybe we should generate types in SDK instead? So you don't have to regenerate them in every app?

krzysztofzuraw: We can - but I think it will the best if we have app-sdk version connected to Saleor version. Then we can generate JSON schema types + GraphQL types as well

PR #453, opened 2025-09-25
Add event for form communication

lkostrowski:


lkostrowski:
> I don't have any comments for code itself, lgtm

I would probably add possibility to subscribe to only specific forms, or single form at a time, so that we don't have handlers that listen to only one form, but get events for all of them.

It's likely not possible because app is mounted on the single context, so you have sinlge form. You can't have open app that is listening to translate product and edit product, because they are different views

PR #509, opened 2026-06-02
Fix widgets resizing API to use `AppBridge.dispatch()` protocol

mirekm:


src/app-bridge/actions.test.ts
```diff
@@ -84,6 +84,16 @@ describe("actions.ts", () => {
     });
   });
 
+  describe("actions.WidgetResize", () => {
+    it("Constructs action with \"widgetResize\" type, random actionId and height payload", () => {
```
lkostrowski, line 88: 1. should `0` be allowed?
2. should `9999999` be allowed?
3. shuld `NaN`, `null`, `"foo"` be allowed?

src/app-bridge/actions.ts
```diff
@@ -172,6 +178,25 @@ function createPopupCloseAction(): PopupClose {
...
+   */
+  height: number;
+};
+
+export type WidgetResize = ActionWithId<"widgetResize", WidgetResizePayload>;
+/**
+ * Reports the app's content height so the Dashboard can resize the widget iframe
+ * to match. Only affects `*_DETAILS_WIDGETS` extensions.
+ */
+function createWidgetResizeAction(payload: WidgetResizePayload): WidgetResize {
```
lkostrowski, line 202: IMO we should store validation here. I mean this is event factory, this is the right place to early inform consumer that something stupid is happening (finite, positive number).

We should validate and either throw immediately or at least warn/error to console, for debugging purposes

IMO its better to throw, because it gives developer earliest possible feedback

src/app-bridge/widget-resize.ts
```diff
@@ -35,41 +19,33 @@ export const measureWidgetHeight = (root: HTMLElement): number => {
...
-  const message: WidgetResizeMessage = {
-    type: WIDGET_RESIZE_MESSAGE,
-    height,
-  };
-
-  window.parent.postMessage(message, "*");
+  // Sizing is best-effort and high-frequency: a Dashboard with no widgetResize action handler sends
+  // no response, so each report rejects after dispatch's 10s timeout. Swallow silently rather than
+  // logging, since these failures are expected and reports are frequent.
+  appBridge.dispatch(actions.WidgetResize({ height })).catch(() => {});
```
lkostrowski: this is fine for backwards compatibility, but we should console.warn it happenend

src/app-bridge/widget-resize.ts
```diff
@@ -35,41 +19,33 @@ export const measureWidgetHeight = (root: HTMLElement): number => {
...
 };
 
 /**
  * Measure a widget root and report its height to the Dashboard.
  *
- * No-op when not embedded in an iframe or when the root is missing.
+ * Pass your widget content root — not `document.body` / `document.documentElement`.
+ * In an iframe those nodes stretch with the iframe and cause incorrect feedback loops.
```
lkostrowski: you can exclude on typescript layer to precent passing document element 
https://html-document.github.io/html-document/docs/class/src/HTMLElement/elements/HTMLDocumentElement.js~HTMLDocumentElement.html

src/app-bridge/widget-resize.ts
```diff
@@ -35,41 +19,33 @@ export const measureWidgetHeight = (root: HTMLElement): number => {
...
  */
-export const postWidgetHeight = (root?: HTMLElement | null): void => {
-  if (SSR || !window.parent || window.parent === window || !root) {
-    return;
-  }
-
-  const height = measureWidgetHeight(root);
-
-  if (!isPositiveFiniteHeight(height)) {
+export const postWidgetHeight = (appBridge: AppBridge, root?: HTMLElement | null): void => {
```
lkostrowski: I dont understand why `root` is optional, so you can *not* pass root and then nothing happens. Its poor design, usually you must go intro source code to understand this 

Instead just make `root `required, otherwise it doesnt make sense to call it

src/app-bridge/widget-resize.ts
```diff
@@ -1,27 +1,11 @@
...
- *
- * Prefer this over `document.body` / `document.documentElement` — in an iframe
- * those nodes stretch with the iframe, which causes incorrect feedback loops.
- */
-export const measureWidgetHeight = (root: HTMLElement): number => {
+/** Measure widget root content height (layout box vs scroll overflow, rounded up). */
+const measureWidgetHeight = (root: HTMLElement): number => {
   if (SSR) {
     return 0;
   }
```
lkostrowski: this is confusing. If element doest exist in SSR, it should not *pretend* to exist (0 is valid html element size). If it doesnt exist, it should return `null`

src/app-bridge/index.ts
```diff
@@ -11,5 +11,9 @@ export * from "./form-payload";
 export * from "./types";
 export * from "./use-dashboard-token";
 export * from "./use-widget-auto-resize";
-export * from "./widget-resize";
+export {
+  type WidgetResizeRootElement,
```
lkostrowski, line 15: why  do you think this should be exported on public api layer?

lkostrowski (approved):

src/app-bridge/index.ts
mirekm: No, it should be not, good point.
I think the public API still needs work. Here's my proposal:

### Exported for the normal path/usage
- `useWidgetAutoResize(root, enabled?)`
- `actions.WidgetResize`

### Exported as escape hatches when you want to control it:
- current `reportWidgetHeight(appBridge, height)` -> `reportWidgetHeight(appBridge, height)`
- current `postWidgetHeight((appBridge, root)` -> `reportWidgetHeightFromElement(appBridge, root)`

PR #511, opened 2026-06-10
Backfill missing units tests

lkostrowski:


stmpn (approved):
I did a one-file-at-a-time pass with Claude. I have reviewed everything he flagged and found nothing that is not a total nitpick. The PR is **gigantic** and mixes so many different places in code, so this is the best I can do here.

PR #527, opened 2026-08-22
Forward Dashboard-owned keyboard shortcuts out of the app iframe

mirekm:
Dashboard global shortcuts (Cmd+K and friends) stop working the moment a user clicks into an app. Apps run in a cross-origin iframe, so every `keydown` goes to the app's document and never reaches the Dashboard's.

This adds a negotiated shortcut channel over the existing App Bridge transport:

1. Dashboard advertises the chords it owns via a new `shortcutsChanged` event.
2. AppBridge stores them and listens for `keydown` in the iframe.
3. On a match it calls `preventDefault()` and dispatches the new `triggerShortcut` action.
4. Dashboard runs the command and moves focus out of the iframe.

Only chords the Dashboard explicitly registered are ever forwarded — typed text never leaves the app.

## Backward compatibility

Forwarding is **on by default** and this is safe: an empty registry matches nothing, so nothing changes until a Dashboard advertises shortcuts. The registry *is* the capability signal, so there's no `dashboardVersion` sniffing.

## Using it — app side

Nothing to do. Any app on this version gets it automatically.

Opt out, or reclaim a chord the app needs for itself:

```ts
new AppBridge({ forwardKeyboardShortcuts: false });

new AppBridge({
  forwardKeyboardShortcuts: {
    shouldForward: (event) => !myEditorIsOpen,
  },
});
```

## Using it — Dashboard side (required, separate repo)

Send the registry right after `handshake`, and again whenever the owned set changes. Full replace, not a delta; send `[]` to revoke.

```ts
appIframe.postMessage({
  type: "shortcutsChanged",
  payload: {
    shortcuts: [
      { id: "commandPalette.open", key: "k", metaKey: true },
      { id: "commandPalette.open", key: "k", ctrlKey: true },
    ],
  },
});
```

Handle the action in `useAppActions`:

```ts
case "triggerShortcut": {
  const command = commands[payload.shortcutId];
  if (!command) return respond(payload.actionId, false);
  command.run();
  return respond(payload.actionId, true);
}
```

Then **focus the resulting UI** (e.g. the palette input). Without an explicit `.focus()`, the iframe keeps focus and typing goes back into the app.

`key` is `KeyboardEvent.key`, matched case-insensitively. Modifiers are exact, so registering Cmd+K does not swallow Cmd+Shift+K. Always respond — silence makes `dispatch` time out after 10s.

## Public API

- `actions.TriggerShortcut({ shortcutId, key, metaKey, ctrlKey, altKey, shiftKey })`
- `shortcutsChanged` event, `DashboardShortcut` type, `appBridgeState.dashboardShortcuts`
- `new AppBridge({ forwardKeyboardShortcuts })` — default `true`
- `matchDashboardShortcut`, `createShortcutForwarder` for non-React hosts
- `AppBridge.destroy()` — detaches the instance from the window

## Also fixed here

- The registry is validated before use. It arrives over `postMessage`, and a malformed entry would otherwise throw on *every keypress* from inside the `keydown` listener.
- `AppBridgeProvider` now destroys a bridge it created on unmount. StrictMode's double-mount left two live bridges on one window — harmless for idempotent events, but it would have run every forwarded shortcut twice.
- `dashboardShortcuts` is optional on `AppBridgeState` so adding it doesn't break consumers that build that type themselves (test doubles).

## Testing

144 app-bridge tests (493 full suite), typecheck clean. Covers the matcher table, repeat/IME suppression, `shouldForward`, malformed payloads, `destroy()`, and the provider lifecycle. The three regression tests were verified to fail without their fix.

⚠️ Merge alongside the Dashboard handler — until it lands, Cmd+K still won't open the palette.


src/app-bridge/actions.ts
```diff
@@ -59,6 +59,15 @@ export const ActionType = {
...
   openPopup: "openPopup",
+  /**
+   * Ask Dashboard to run a shortcut it previously advertised via
+   * `shortcutsChanged`. AppBridge dispatches this automatically when the user
+   * presses a registered chord inside the iframe.
+   *
+   * Only has an effect on Dashboard versions that handle the `triggerShortcut`
+   * action type and send `shortcutsChanged`.
+   */
+  triggerShortcut: "triggerShortcut",
```
lkostrowski, line 75: having it registered here means we make it public api. EAch time app developer uses `actions.` typescript will suggest `TriggerShortcut`. But should it? 
1. CMD+K is supposed to listen in the background, without explicit knowledge 
2. Other shortcuts are not allowed and we shouldn't incentivize to act on behalf user (if we want app to open K-bar we should register `OpenCommandPalette()` command)

So IMO this action should not be exported within `actions` - maybe export it as separate symbol/file. Even better if you export it as `InternalTriggerShortcut` or something like this for even better DX

src/app-bridge/app-bridge-state.ts
```diff
@@ -36,6 +36,15 @@ export type AppBridgeState = {
...
   appParams?: unknown;
+  /**
+   * Keyboard shortcuts the Dashboard currently owns. Replaced in full by each
+   * `shortcutsChanged` event. Empty until the Dashboard advertises any — older
+   * Dashboards never send the event, so forwarding stays a no-op.
+   *
+   * Optional so that adding it doesn't break consumers that build an
+   * `AppBridgeState` themselves (e.g. test doubles). AppBridge always sets it.
+   */
+  dashboardShortcuts?: DashboardShortcut[];
```
lkostrowski, line 47: this is also internal API so I would at least adjust description, and best if we don't even store it publically, eg

```
export const internalDashboardActions = new Symbol()
appBridgeState[internalDashboardActions] = [...]

// access only internally via symbol
```

or try with private class propery

lkostrowski (changes requested):

src/app-bridge/actions.ts
mirekm: Agreed. `triggerShortcut` is internal plumbing, not an app API.

Removed the `the actions.TriggerShortcut` helper from actions. It now lives next to the keyboard forwarder and is not re-exported. Wire type stays so Dashboard can implement the handler.

Would that work?

lkostrowski (approved):
lets try

